PR #2, opened 2015-07-01
Bump fetch to 0.9.0

topaxi:
For some reason, npm installed 0.7.0 (I assume "^0.7.0" should install 0.9.0 too?), which failed my code because i used `response.ok`... It works now with 0.9.0, I bump here so I don't have to update all my package.json files in the future :)


stefanpenner:
^ drifts `0.7.x` and `1.x.x` basically, the first non-zero version stays fixed.

stefanpenner:
release as `v1.2.1`

PR #5, opened 2015-09-28
fix form data and blob compat

calvinmetcalf:
FormData, Blob, and FileReader need to be on the self object, otherwise thinks they aren't supported

### DIFF
--- a/tests/acceptance/root-test.js
+++ b/tests/acceptance/root-test.js
@@ -70,7 +70,7 @@ test('posting a form', function(assert) {
   });
   var form = new window.FormData();
   form.append('foo', 'bar');
-  fetch('/upload', {
+  return fetch('/upload', {
     method: 'post',
     body: form
   }).then(function (res) {
@@ -78,7 +78,5 @@ test('posting a form', function(assert) {
     return res.json();
   }).then(function (data) {
     assert.equal(data.name, 'World');
-  }).catch(function (e) {
-    assert.ok(false, e.stack);
   });
 });
--- a/tests/acceptance/root-test.js
+++ b/tests/acceptance/root-test.js
@@ -5,6 +5,7 @@ import {
 } from 'qunit';
 import startApp from 'dummy/tests/helpers/start-app';
 import Pretender from 'pretender';
+import fetch from 'fetch';
 var application;
 var server;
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
@calvinmetcalf window isn't always available.

calvinmetcalf:
ok now it should work in web workers, will get a test next

stefanpenner:
@calvinmetcalf LGTM, thanks for catching this one!

calvinmetcalf:
I'd be happy to add a test for this, but I'm not familiar enough with how the ember tests are setup to add the first test to this repo.

stefanpenner:
some places to look:

* test likely could go here: https://github.com/stefanpenner/ember-fetch/blob/master/tests/acceptance/root-test.js
* if you need to fake the underlying XHR, we use [pretender](https://github.com/pretenderjs/pretender)

if you have other specific questions i can try to answer them

calvinmetcalf:
done, though I do need to point out that qunit is terrible for async tests (also in general)

stefanpenner:
@calvinmetcalf you can actually just return the promise. And qunit will automatically go async, and also assert fulfillment and assert against rejection.  Should cleanup stuff and make qunit more tolerable.

calvinmetcalf:
well that certainly is an improvement over having to call stop

tests/acceptance/root-test.js
```diff
@@ -37,3 +38,45 @@ test('visiting /', function(assert) {
...
+test('posting a string', function(assert) {
+  server.post('/upload', function(req) {
+    assert.equal(req.requestBody, 'foo');
+    return [
+      200,
+      { 'Content-Type': 'text/json'},
+      JSON.stringify({ name: 'World' })
+    ];
+  });
+  fetch('/upload', {
```
stefanpenner: Missing return?

tests/acceptance/root-test.js
```diff
@@ -37,3 +38,45 @@ test('visiting /', function(assert) {
...
+    method: 'post',
+    body: 'foo'
+  }).then(function (res) {
+    assert.equal(res.status, 200);
+    return res.json();
+  }).then(function (data) {
+    assert.equal(data.name, 'World');
+  }).catch(function (e) {
+    assert(false, e);
```
stefanpenner: This won't be needed once return above is added

stefanpenner:
Some last few comments. Once addressed, we can merge and I'll release

calvinmetcalf:
fixed

PR #6, opened 2015-09-29
use globals if avialable

calvinmetcalf:
not sure how to test this as this is being set once on module load, do I have access to the resolver inside the test module ?

### DIFF
--- a/assets/module-template.js.t
+++ b/assets/module-template.js.t
@@ -1,25 +1,32 @@
 (function (global) {
   define('fetch', [ 'ember', 'exports' ], function(Ember, self) {
     'use strict';
+    if (global.fetch) {
+      self['default'] = self.fetch = global.fetch;
+      self.Response = global.Response;
+      self.Request = global.Request;
+      self.Headers = global.Headers;
+      return;
+    }
     var Promise = Ember['default'].RSVP.Promise;
     if (global.FormData) {
       self.FormData = global.FormData;
     }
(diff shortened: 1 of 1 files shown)


stefanpenner:
> user globals if avialable

just double checking, you meant `use` not `user` right?

> do I have access to the resolver inside the test module ?

sorry, I don't quite understand

calvinmetcalf:
- yes typo
- so the way I'd go about testing this especially in phantom, would be I'd put a fake fetch object on window an see if it was what I got when I imported fetch, but you can't do that with ES6 module syntax as they will always load before the current module is run so my guess as to the best way to test this would be to access the module resolver somehow...

stefanpenner:
So one approach, would be to change: https://github.com/stefanpenner/ember-fetch/blob/master/assets/module-template.js.t#L5-L13 to instead use `defineProperty` so access to self.Header is always actually being fetched from global.Header

This would allow patching during tests, and likely more resilient to other related polyfils being added.

calvinmetcalf:
I though the whole point of this was so it wouldn't pollute global scope?

stefanpenner:
I think that continues to be true. I can take a closer look after lunch

calvinmetcalf:
In which case I may be confusing what you meant, one idea could be to to not export the self object that the shim binds too but to put getters on the export object which dynamically grabs global.fetch/Request/whatever or the shim one.

PR #11, opened 2016-01-31
Upgrade deps

aquamme:
It looks like this wrapper can get out of date due to the version selector in package.json - not sure if you want to revise that. I just upgraded ember-cli and related dependencies, as well as the polyfill to version 0.11.0.


bower.json
```diff
@@ -1,17 +1,20 @@
...
-    "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
-    "ember-data": "1.0.0-beta.16.1",
-    "ember-load-initializers": "ember-cli/ember-load-initializers#0.1.4",
-    "ember-qunit": "0.3.1",
-    "ember-qunit-notifications": "0.0.7",
-    "ember-resolver": "~0.1.15",
-    "jquery": "^1.11.1",
-    "loader.js": "ember-cli/loader.js#3.2.0",
-    "qunit": "~1.17.1",
+    "ember": "components/ember#release",
```
stefanpenner: any reason why this isn't a fixed version? ember-try should handle release channels

aquamme: Only because it's what `ember init` output. I can change it to `2.3.x` or whatever you recommend.

stefanpenner: nah thats ok

Issue #19, opened 2017-03-15
self is not window => always polyfilling


craigbilner:
Hi,

I'm having a bit of an issue with both this lib and [ember-network](https://github.com/tomdale/ember-network). They both wrap the `whatwg-fetch` in an IIFE which tests `self.fetch` _but_ `self` is not `window`.

It lives inside this IIFE:

```
(function (global) {
  define('fetch', ['ember', 'exports'], function(Ember, self) {
...
```

at this point `self` is a pojo which means `fetch` is **always polyfilled**. I'm not overly familiar with the `reification` process but I can't quite see how `exports` maps to something that could be tested for fetch?

Is this some sort of build issue my side or should the variable `self` be renamed so it's not shadowed by the `whatwg-fetch` test?

[raised a speculative PR #20]

tchak:
@craigbilner this is by design for this addon. We can not reasonably use native `fetch` until `pretender` supports `fetch`.

stefanpenner:
This is also so that features that are not polyfillable are not accidentally leaked forward.

I'm not apposed to changing this, or making it configurable, but allowing fallback as it stands currently poses a hefty risk.

craigbilner:
thanks @tchak @stefanpenner 

Could we make it configurable and/or update the README to this effect? Others may think they're getting fetch when they're not and may not use Pretender?

stefanpenner:
Yup would accept pr doing so

PR #20, opened 2017-03-15
Change self to exp

craigbilner:
Decorate the export object with "real self", which is either native `fetch` or the polyfilled one, which has used "real self" to test for native.


tchak:
see #19 for discussion

PR #24, opened 2017-04-05
see if this makes windows happy

stefanpenner:


stefanpenner:
@tchak windows CI is now happy

PR #25, opened 2017-04-06
Allow automatic "waiting" in tests.

rwjblue:
This has been fundamentally supported when using $.ajax based requests for quite some time (since at least Ember 1.8ish).

This change brings testing parity to ember-fetch...

### DIFF
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -14,7 +14,21 @@
 
     <%= moduleBody %>
 
-    self['default'] = self.fetch;
+    var pending = 0;
+    if (Ember.Test) {
+      Ember.Test.registerWaiter(function() { return pending === 0; });
+    }
+
+    function decrement(result) {
+      pending--;
+      return result;
+    }
+
+    self['default'] = function() {
+      pending++;
+
+      return self.fetch.apply(self, arguments).then(decrement, decrement);
+    };
     self['Headers'] = self.Headers;
     self['Request'] = self.Request;
     self['Response'] = self.Response;


cibernox:
@rwjblue check this first: https://github.com/emberjs/ember-test-helpers/pull/202

rwjblue:
Yeah, that PR looks nice, but it doesn't really help us here AFAICT.  Main reasons:

1. `ember-test-helpers/wait` implementation is only used for unit and integration tests (not acceptance tests).  The acceptance test infrastructure is (for now) still in emberjs/ember.js itself.
2. Things downstream that run in production cannot easily know if `ember-test-helpers` module is present so a straight `import { incrementPendingRequestCount } from 'ember-test-helpers'` cannot be done, because when `ember-test-helpers` is not present an error will be thrown by the loader.
3. I think a promise based API would still be nice to add support for.  (this is really just an opinion thing though)

stefanpenner:
looks good, we should add 1 acceptance test in the dummy app so we dont regress in the future.

cibernox:
The problem with approach is that it adds waiters instead of ajax requests. `wait` accepts an options hash to skip waiting for a few things. Passing `wait({ waitForAjax: false })` should affect $.ajax and `fetch` as both network requests and should be grouped.

However I never saw that feature in use, probably no one would ever notice.

If we wanted to provide helpers (and a nicer promise-based API too), how would you make that import go await in production? Babel transforms? 

I could make the same refactor in Ember itself.

cibernox:
I've discovered that my the approach in my PR was flawed and probably this is too. 

A numeric counter is not enough because if the user doesn't wait it is possible for a not-awaited fetch requests to end in the next test, incorrectly decrementing the counter.

We should keep an array of pending promises and when they are fulfilled, slice them from the array. The timer needs to wait until the array has zero length.

cibernox:
I take it back. Since this approach doesn't reset the counter of requests between tests, this approach is safe. The worst that can happen is that a tests has to wait a bit longer because a pending promise from the previous test.

cibernox:
I used this fork and it's better, but not quite there yet. Before test failed, not they pass just ~50% of the time: https://github.com/cibernox/ember-native-dom-helpers/pull/48

My only guess is that there is a race condition. The **only** place where I can see that happening, and I hope I'm wrong because I won't see an easy solution.

My theory is that the promise from `fetch` resolves when the response starts comming back, but the resolved value (a `Response` object) might still be transmitting. You await for this stream of data by using `reponse.json()` or `response.text()`, all of which return another promise.

I suspect that if the data transfer is not fast enough, the waiter is going to continue and thats why tests fail intermittently.

I think that for fixing that we need to wait for the data transfer to finish (only in testing) and return a cloned response.

Untested pseudo javascript:

```js
if (detectTestingEnvironment) {
  var pending = 0;
  if (Ember.default.Test) {
    Ember.default.Test.registerWaiter(function() { return pending === 0; });
  }

  function decrement(result) {
    pending--;
    return result;
  }

  self['default'] = function() {
    pending++;
    self.fetch.apply(self, arguments).then((response) => {
      response.clone().blob().then(decrement,decrement);
      return response;
    }, decrement);
  }
} else {
  self['default'] = self.fetch;
}

```

cibernox:
@stefanpenner Friendly ping

stefanpenner:
@cibernox thanks for the ping, it looks like the acceptance test (to ensure this works) is still missing. If that is added, I am good to 🚢

stefanpenner:
> @cibernox thanks for the ping, it looks like the acceptance test (to ensure this works) is still missing. If that is added, I am good to 🚢

ping, if we can get a test here. I'll merge and release!

cibernox:
C**p, I forgot about this 😞 

I'll try to get this done at some point today.

samselikoff:
@cibernox Should this still work out of the box today? Anything special needed to wire things up? I'm having trouble getting my test to wait. (I'm seeing the `registerWaiter` call you added here, but the actual waiter that checks for `pending === 0` never seems to be invoked.)

viniciussbs:
@samselikoff are you still having problems with it? I'm facing the same problem. I guess those issues are related: https://github.com/emberjs/ember-test-helpers/issues/477 and https://github.com/emberjs/ember-test-helpers/issues/347.

samselikoff:
I haven't checked back on this, it was for some older work

walter:
I'm moving an `ic-ajax` based service to be `ember-fetch` (wrapper around `fetch` calls).  Got it working except for running into issues with tests not waiting. I've sprinkled a crap-ton of `await settled()` calls in my tests as a temporary workaround. So yeah, seems to still be a problem.

I just confirmed the problem in the `6.3.1` release.

walter:
Also note using latest `ember-qunit` (`4.2.0`) and `ember-test-helpers` (`1.1.0`).

rwjblue:
It’s pretty unlikely that commenting here (long dead PR) will result in help... 

Would you mind filing a new issue (hopefully with a reproduction) so we can try to figure out what’s going on?

walter:
I'm finishing up my last day of work before mid-January. Unlikely I'll be able to create a new issue with a reproduction before I sign-off for the break. If @viniciussbs doesn't beat me to it, I'll put something together when I return.

viniciussbs:
My issue was on https://github.com/bgentry/ember-apollo-client/issues/192.

PR #27, opened 2017-04-12
Avoid hidden filename collision with ember-network

### LINKED ISSUE #26: Import `fetch from 'fetch'` not working w/ 1.4.0
Doing `import fetch from 'fetch'` throws a `ember.debug.js:18015 Error: Could not find module 'fetch'` on 1.4.0. Works fine on 1.3.0

- Ember-Cli 2.12.1 w/ latest dependencies
- Node 7.4.0

ef4:
This fixes #26. ember-network provides a file named vendor/fetch.js, which can overwrite the one that comes from this addon.

Why would you have both ember-network and ember-fetch? In my case it was because ember-simple-auth was pulling in ember-network without my realizing.

In general I think addons should try to use their own name when placing things in vendor, so they don't compete over common names.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,7 @@ module.exports = {
    * from our `vendor` tree into the final built app.
    */
   included: function(app) {
-    app.import('vendor/fetch.js', {
+    app.import('vendor/ember-fetch.js', {
       exports: {
         default: [
           'default',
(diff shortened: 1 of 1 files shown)


stefanpenner:
> Why would you have both ember-network and ember-fetch? In my case it was because ember-simple-auth was pulling in ember-network without my realizing.

Not sure why either, as ember-network is a copy/pasta of this project....

stefanpenner:
`released as vv1.4.1 🎉`

stefanpenner:
thanks @ef4 !

PR #31, opened 2017-05-25
Ember Data Adapter Mixin to use fetch

nlfurniss:
I would have liked to do some more testing of the adapter methods, let's discuss how best to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,24 @@ export default Ember.Route.extend({
 });
 ```
 
+### Use with Ember Data
+To have Ember Data utilize `fetch` instead of jQuery.ajax to make calls to your backend, extend your project's `application` adapter with the `adapter-fetch` mixin.
+
+```js
+// app/adapters/application.js
+import DS from 'ember-data';
+import AdapterFetch from 'ember-fetch/mixins/adapter-fetch';
+
+export default RESTAdapter.extend(AdapterFetch, {
+  ...
+});
+```
+
 further docs: https://github.com/github/fetch
 
 ### Browser Support
 
-* evergreen / IE9+ / Safari 6.1+ https://github.com/github/fetch#browser-support
+* evergreen / IE10+ / Safari 6.1+ https://github.com/github/fetch#browser-support
 
 
 ### does this replace ic-ajax?
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -130,7 +130,9 @@ export default Ember.Mixin.create({
       })
       .then((response) => {
         if (response.ok) {
-          const bodyPromise = response.json();
+          // We want to check that there is a body in the response, otherwise JSON.parse will throw an error.
+          // Instead, we'll let Ember Data handle an empty response.
+          const bodyPromise = response.headers.get('content-length') ? response.json() : {};
           return this.ajaxSuccess(response, bodyPromise, requestData);
         }
         throw this.ajaxError(null, response, requestData);
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -81,13 +81,16 @@ export function headersToObject(headers) {
  * @param {Object} options
  */
 export function mungOptionsForFetch(_options, adapter) {
-  const options = (assign || merge)({
+  // This allows this mixin to be backward compatible with Ember < 2.5.
+  const combineObjs = (assign || merge);
+  const options = combineObjs({
     credentials: 'same-origin',
   }, _options);
 
   let adapterHeaders = adapter.get('headers');
   if (adapterHeaders) {
-    options.headers = assign({}, options.headers || {}, adapterHeaders);
+    // This double use of `combineObjs` is necessary because `merge` only accepts two arguments.
+    options.headers = combineObjs(combineObjs({}, options.headers || {}), adapterHeaders);
   }
   options.method = options.type;
 
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -7,6 +7,8 @@ const {
   RSVP
 } = Ember;
 
+const RBRACKET = /\[\]$/;
+
 /**
  * Helper function that turns the data/body of a request into a query param string.
  * This is directly copied from jQuery.param.
@@ -14,22 +16,16 @@ const {
  * @returns {String}
  */
 export function serialiazeQueryParams(queryParamsObject) {
-  var s = [], rbracket = /\[\]$/;
-  function isArray(obj) {
-    return Object.prototype.toString.call(obj) === '[object Array]';
-  }
-  function add(k, v) {
-    v = typeof v === 'function' ? v() : v === null ? '' : v === undefined ? '' : v;
-    s[s.length] = encodeURIComponent(k) + '=' + encodeURIComponent(v);
-  }
+  var s = [];
+
   function buildParams(prefix, obj) {
     var i, len, key;
 
     if (prefix) {
-      if (isArray(obj)) {
+      if (Array.isArray(obj)) {
         for (i = 0, len = obj.length; i < len; i++) {
-          if (rbracket.test(prefix)) {
-            add(prefix, obj[i]);
+          if (RBRACKET.test(prefix)) {
+            add(s, prefix, obj[i]);
           } else {
             buildParams(prefix + '[' + (typeof obj[i] === 'object' ? i : '') + ']', obj[i]);
           }
@@ -39,11 +35,11 @@ export function serialiazeQueryParams(queryParamsObject) {
           buildParams(prefix + '[' + key + ']', obj[key]);
         }
       } else {
-        add(prefix, obj);
+        add(s, prefix, obj);
       }
-    } else if (isArray(obj)) {
+    } else if (Array.isArray(obj)) {
       for (i = 0, len = obj.length; i < len; i++) {
-        add(obj[i].name, obj[i].value);
+        add(s, obj[i].name, obj[i].value);
       }
     } else {
       for (key in obj) {
@@ -54,6 +50,17 @@ export function serialiazeQueryParams(queryParamsObject) {
   }
 
   return buildParams('', queryParamsObject).join('&').replace(/%20/g, '+');
+}
+
+/**
+ * Part of the `serialiazeQueryParams` helper function.
+ * @param {Array} s
+ * @param {String} k
+ * @param {String} v
+ */
+function add(s, k, v) {
+  v = typeof v === 'function' ? v() : v === null ? '' : v === undefined ? '' : v;
+  s[s.length] = `${encodeURIComponent(k)}=${encodeURIComponent(v)}`;
 }
 
 /**
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -1,0 +1,189 @@
+import Ember from 'ember';
+import fetch from 'fetch';
+
+const {
+  assign,
+  merge,
+  RSVP
+} = Ember;
+
+/**
+ * Helper function that turns the data/body of a request into a query param string.
+ * This is directly copied from jQuery.param.
+ * @param {Object} queryParamsObject
+ * @returns {String}
+ */
+export function serialiazeQueryParams(queryParamsObject) {
+  var s = [], rbracket = /\[\]$/;
+  function isArray(obj) {
+    return Object.prototype.toString.call(obj) === '[object Array]';
+  }
+  function add(k, v) {
+    v = typeof v === 'function' ? v() : v === null ? '' : v === undefined ? '' : v;
+    s[s.length] = encodeURIComponent(k) + '=' + encodeURIComponent(v);
+  }
+  function buildParams(prefix, obj) {
+    var i, len, key;
+
+    if (prefix) {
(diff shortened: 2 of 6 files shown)


kratiahuja:
@nlfurniss how will this work with things like bigpipe?

nlfurniss:
Yup it works fine with bigpipe (used v-web to test this). The only thing I need to confirm is that the proper tracking headers are attached to requests BPR sends. Have sent Mark and Rajul an email to figure out how to confirm.

stefanpenner:
I like this as a stepping stone for ember-data users to use ember-fetch 👍

README.md
```diff
@@ -28,11 +28,14 @@ export default Ember.Route.extend({
 });
 ```
 
+### Use with Ember Data
+To have Ember Data utilize `fetch` instead of jQuery.ajax to make calls to your backend, extend your project's `application` adapter with the `adapter-fetch` mixin.
```
stefanpenner: lets give a quick code example.

README.md
```diff
@@ -28,11 +28,14 @@ export default Ember.Route.extend({
...
+
 further docs: https://github.com/github/fetch
 
 ### Browser Support
 
-* evergreen / IE9+ / Safari 6.1+ https://github.com/github/fetch#browser-support
+* evergreen / IE10+ / Safari 6.1+ https://github.com/github/fetch#browser-support
```
stefanpenner: ah is the polyfil IE10+ only now?

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,147 @@
...
+    headerObject[key] = value;
+  });
+  return headerObject;
+}
+/**
+ * Helper function that translates the options passed to `jQuery.ajax` into a format that `fetch` expects.
+ * @param {Object} options
+ */
+export function mungOptionsForFetch(options) {
+  const _options = assign({
```
stefanpenner: tiny nit pick, but can we make the param be `_options` but the one used throughout the function `option` I think this reads nicer.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,147 @@
...
+    return this._ajaxRequest(hash)
+      .then((response) => {
+        if (response.ok) {
+          const bodyPromise = response.json();
+          return this.ajaxSuccess(response, bodyPromise, requestData);
+        }
+        throw this.ajaxError(null, response, requestData);
+      })
+      .catch((error, response, requestData) => {
+        throw this.ajaxError(error, response, requestData);
```
stefanpenner: As implemented this `catch` will also catch all errors thrown in the `then` which precedes it. This includes its own `this.ajaxError`  That seems funky.

I believe simply re-ordering the `then` and the `catch` would address this issue.

```js
this._ajaxRequest(hash)
  .catch((error, response, requestData) => { 
    throw.this.ajaxError(error, response, requestData);
  }).then(response => {
    if (response.ok) {
      const bodyPromise = response.json();
      return this.ajaxSuccess(response, bodyPromise, requestData);
    }
    throw this.ajaxError(null, response, requestData);
  });
```

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,147 @@
...
+    return bodyPromise.then((body) => {
+      let returnResponse;
+      try {
+        returnResponse = this.handleResponse(
+          response.status,
+          headersToObject(headerString),
+          body,
+          requestData
+        );
+      } catch (error) {
```
stefanpenner: this `try/catch` is redundant and can be removed. the caught error will already be caught be the `then` handler it is being called in.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,147 @@
...
+      try {
+        const headerString = headersToObject(response.headers);
+        returnedError = this.handleResponse(
+          response.status,
+          headersToObject(headerString),
+          this.parseErrorResponse(response.statusText) || error,
+          requestData
+        );
+      } catch (e) {
+        returnedError = e;
```
stefanpenner: any downside to letting this just `throw` ?

README.md
nlfurniss: Yup, according to their readme: https://github.com/github/fetch#browser-support

stefanpenner: thanks

nlfurniss:
Updated with @stefanpenner's suggestions

nlfurniss:
so phantom doesn’t have `Map` or `Headers`, so my test of the `headersToObject` helper is failing.

do you want to shim Map/Headers, leave out the test, other?

tchak:
@nlfurniss polyfilled `Headers` should be available as `import { Headers } from 'ember-fetch';`

cibernox:
@nlfurniss This is fantastic 👏 

Counting the seconds to use it.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,142 @@
...
+  const options = assign({
+    credentials: 'same-origin',
+  }, _options);
+
+  options.method = options.type;
+
+  // GET and HEAD requests can't have a `body`
+  if (options.data && Object.keys(options.data).length) {
+    if (options.method === 'GET' || options.method === 'HEAD') {
+      options.url += `?${serialiazeQueryParams(_options.data)}`;
```
cibernox: I found one error. This does not serialize nested objects properly.

Example:

`serialiazeQueryParams(_options.data)` with this implementation generates `page=%5Bobject%20Object%5D`.

The expected queryParams that is generated by `$.ajax` is `page%5Bnumber%5D=1&page%5Bsize%5D=3`

cibernox: I've been playing in the console, this snippet seems to work well for arbitrary nested objects:
```
function serialiazeQueryParams(queryParamsObject, prefix = null) {
  return Object.keys(queryParamsObject).sort().map((key) => {
    let value = queryParamsObject[key];
    if (prefix) {
      key = `${prefix}[${key}]`;
    }
    if (value !== null && typeof value === 'object') {
      return serialiazeQueryParams(value, key);
    } else {
      return `${encodeURIComponent(key)}=${encodeURIComponent(value)}`;
    }
  }).join('&');
}
```

I haven't checked how arrays should work with arrays.

For what is worth, this is the implementation of `$.param`, which is the thing used by `$.ajax`: https://github.com/jquery/jquery/blob/master/src/serialize.js#L17-L90

nlfurniss: We spoke offline and this functionality has been added, along with tests

README.md
nlfurniss: Done!

cibernox:
@nlfurniss Why was this closed?

nlfurniss:
an accident of those history fixes I did :-\

But it's back!

cibernox:
@nlfurniss You scared the s**** out of me.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+        throw this.ajaxError(null, response, requestData);
+      });
+  },
+  /**
+   * Overrides the `_ajaxRequest` method to use `fetch` instead of jQuery.ajax
+   * @param {Object} options
+   * @override
+   */
+  _ajaxRequest(options) {
+    const _options = mungOptionsForFetch(options);
```
cibernox: Added a PR for this.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+
+    const hash = this.ajaxOptions(url, type, options);
+
+    return this._ajaxRequest(hash)
+      .catch((error, response, requestData) => {
+        throw this.ajaxError(error, response, requestData);
+      })
+      .then((response) => {
+        if (response.ok) {
+          const bodyPromise = response.json();
```
cibernox: This line also fails if the server responds with a `204 No content`, since JSON.parse throws an error trying to parse an empty response.

addon/mixins/adapter-fetch.js
nlfurniss: @stefanpenner, how do you feel about adding ED to this addon?

[Pull request](https://github.com/nlfurniss/ember-fetch/pull/2)

addon/mixins/adapter-fetch.js
nlfurniss: Is using the `Content-Length` header reliable enough? or is 204 an edge case?

```
let bodyPromise;
if (response.ok && response.status !== 204) {
  bodyPromise = response.json();
} else {
  bodyPromise = {};
}
...
```

cibernox: No idea. I've asked in slack what the right way of handing 204s is.

cibernox:
After some debugging I found that this approach doesn't work if you also run your app in fastboot. I opened this issue in there: https://github.com/ember-fastboot/ember-cli-fastboot/issues/419#issuecomment-306638366

The **tl;dr** is that fastboot has an initializer that also replaces the `_ajaxRequest` method on all adapters to use `node-ajax`. This mixin and that initializer are essentially fighting.

The solution I see is to put this mixing in a standalone addon, not in here, so that addon can explicitly initialize itself after `ember-cli-fastboot` and trump over it.

Another approach if the people in ember-cli-fastboot agree would be to move the ember-data compatibility layer elsewhere. Perhaps to ember-data itself, perhaps to an `ember-data-fastboot-adapter` people has to install.

stefanpenner:
> The tl;dr is that fastboot has an initializer that also replaces the _ajaxRequest method on all adapters to use node-ajax. This mixin and that initializer are essentially fighting.

We will need to tackle this independent of this issue. I believe fastboot XHR/FETCH interactions needs some love.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+  function add(k, v) {
+    v = typeof v === 'function' ? v() : v === null ? '' : v === undefined ? '' : v;
+    s[s.length] = encodeURIComponent(k) + '=' + encodeURIComponent(v);
+  }
+  function buildParams(prefix, obj) {
+    var i, len, key;
+
+    if (prefix) {
+      if (isArray(obj)) {
```
stefanpenner: `Array.isArray`

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * Helper function that turns the data/body of a request into a query param string.
+ * This is directly copied from jQuery.param.
+ * @param {Object} queryParamsObject
+ * @returns {String}
+ */
+export function serialiazeQueryParams(queryParamsObject) {
+  var s = [], rbracket = /\[\]$/;
+  function isArray(obj) {
```
stefanpenner: we don't need this helper, `Array.isArray` is available in our target browsers and is also faster.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+    return Object.prototype.toString.call(obj) === '[object Array]';
+  }
+  function add(k, v) {
```
stefanpenner: lets not redefine this method on each invocation of `serialiazeQueryParams`, instead hoist it up.

addon/mixins/adapter-fetch.js
```diff
@@ -0,0 +1,189 @@
...
+} = Ember;
+
+/**
+ * Helper function that turns the data/body of a request into a query param string.
+ * This is directly copied from jQuery.param.
+ * @param {Object} queryParamsObject
+ * @returns {String}
+ */
+export function serialiazeQueryParams(queryParamsObject) {
+  var s = [], rbracket = /\[\]$/;
```
stefanpenner: rbracket should likely be `const RRACKET = /...` at the top of the module (no need to redefine it.

stefanpenner:
@cibernox you mentioned the fastboot thing, do you feel that is a blocker here?

cibernox:
@stefanpenner I'd say that the inability if use ember-fetch (this mixin only, the rest works well) along with fastboot might be a problem for ppl building PWAs without jquery. 

TBH, I think that in this case fastboot's approach is a bit intrusive.

kratiahuja:
> TBH, I think that in this case fastboot's approach is a bit intrusive.

Why is fastboot's approach a bit intrusive? I think the problem is the union with `ember-data`. If you aren't using `ember-data` this would work out of the box (even in fastboot). If you chose to use `ember-data`, then ofcourse you will need to provide a way to override the behavior in fastboot like you are doing here with the mixin.

stefanpenner:
> Why is fastboot's approach a bit intrusive?

I think fastboot may want to not monkey patch ember-data, but rather provide an XMLHTTPRequest and Fetch global. I think that would be less intrusive, and work with more libraries (pretender/mirage etc).

stefanpenner:
released as v3.0.0 🎉

issue for fastboot + ED compat: https://github.com/stefanpenner/ember-fetch/issues/38

kratiahuja:
> I think fastboot may want to not monkey patch ember-data, but rather provide an XMLHTTPRequest and Fetch global. I think that would be less intrusive, and work with more libraries (pretender/mirage etc). This idea needs more investigation/exploration etc...

Agreed the monkey patching is bad. but I also think `ember-data` should be providing an network interface for fastboot or anyone else to define. This way even `ember-fetch` won't need to define a mixin but just defines the interface for `ember data`.

PR #34, opened 2017-06-07
[TYPESCRIPT]: added early index.d.ts for type defs

toranb:
A simple type def file I'm using to build apps with ember -thought to see if @stefanpenner and @krisselden agree on what I've got here


toranb:
I've been using this with GET and POST (with params) all week without issue. At worst we could merge this and iterate (with additional type information as needed/ adoption takes off at some point in the future) - thoughts?

PR #37, opened 2017-06-15
[FIXES #35] ensure rejections remain rejections in tests

### LINKED ISSUE #35: Fetch promises always resolving
Hello,
I'm using Ember 2.9, and it appears that the promise returned from` fetch` always resolves, even if there is an error.

I think it's down to this line: https://github.com/stefanpenner/ember-fetch/blob/c2cbcb97a1dfd6605b02c7ea4e8627d5cea5add6/assets/browser-fetch.js.t#L20

Returning the value rather than throwing it seems to trigger the `success` route in the Ember `invokeCallback` function.

Am I going crazy or is this an issue?!

stefanpenner:
my guess at fixing #35, but I wasn't able to quickly get a failing test so I am not sure I understood the problem correctly.

@orf  / @cibernox / @rwjblue a failing test would make this easier to quickly address.

### DIFF
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -31,7 +31,10 @@
         return self.fetch.apply(self, arguments).then(function(response){
           response.clone().blob().then(decrement, decrement);
           return response;
-        }, decrement);
+        }, function(reason) {
+          decrement(reson);
+          throw reason;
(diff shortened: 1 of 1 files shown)


rwjblue:
closed in favor of #44

PR #39, opened 2017-06-16
Empty response should yield Promise not empty object

nlfurniss:
Without this change, the code in `ajaxSuccess` will fail:

```
ajaxSuccess(response, bodyPromise, requestData) {
    const headersObject = headersToObject(response.headers);

    return bodyPromise.then((body) => {   // {} has no `then`
```

@stefanpenner @cibernox

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -117,9 +117,11 @@ export function mungOptionsForFetch(_options, adapter) {
  */
 export function determineBodyPromise(response) {
   let bodyPromise;
-  const contentLength = response.headers.get('content-length');
 
-  if (contentLength && Number(contentLength)) {
+  // Content-Length is returned as a string or null, so we convert to a number.
+  const contentLength = Number(response.headers.get('content-length'));
+
+  if (contentLength > 0) {
     bodyPromise = response.json();
   } else {
     bodyPromise = RSVP.Promise.resolve({});
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -109,6 +109,23 @@ export function mungOptionsForFetch(_options, adapter) {
 
   return options;
 }
+/**
+ * Function checks that there is a body in the response. Otherwise JSON.parse will throw an error.
+ * If there is not, return an empty object and let Ember data handle the empty response
+ * @param {Response} response
+ * @returns {Promise}
+ */
+export function determineBodyPromise(response) {
+  let bodyPromise;
+  const contentLength = response.headers.get('content-length');
+
+  if (contentLength && parseInt(contentLength)) {
+    bodyPromise = response.json();
+  } else {
+    bodyPromise = RSVP.Promise.resolve({});
+  }
+  return bodyPromise;
+}
 
 export default Ember.Mixin.create({
   /**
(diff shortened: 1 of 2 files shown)


addon/mixins/adapter-fetch.js
```diff
@@ -133,7 +133,12 @@ export default Ember.Mixin.create({
         if (response.ok) {
           // We want to check that there is a body in the response, otherwise JSON.parse will throw an error.
           // Instead, we'll let Ember Data handle an empty response.
-          const bodyPromise = response.headers.get('content-length') ? response.json() : {};
+          let bodyPromise;
+          if (response.headers.get('content-length')) {
+            bodyPromise = response.json();
+          } else {
+            bodyPromise = new RSVP.Promise((resolve) => resolve({}));
```
stefanpenner: Could also be expressed as: `RSVP.Promise.resolve({})`

addon/mixins/adapter-fetch.js
```diff
@@ -109,6 +109,23 @@ export function mungOptionsForFetch(_options, adapter) {
...
+ * Function checks that there is a body in the response. Otherwise JSON.parse will throw an error.
+ * If there is not, return an empty object and let Ember data handle the empty response
+ * @param {Response} response
+ * @returns {Promise}
+ */
+export function determineBodyPromise(response) {
+  let bodyPromise;
+  const contentLength = response.headers.get('content-length');
+
+  if (contentLength && parseInt(contentLength)) {
```
stefanpenner: I believe `Number` constructor ends up having less caveats then parseInt. 

For example, Number is less forgiving when given junk:

```js
parseInt(' 10dd1') // => 10
Number(' 10dd1') // => NaN
```

thoughts?

tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -172,3 +176,43 @@ test('serialiazeQueryParams turns deeply nested objects into queryParams like $.
...
+    {
+      status: 200,
+      headers: new Headers({
+        'Content-Length': 0
+      })
+    }
+  );
+  const bodyPromise = determineBodyPromise(response);
+
+  bodyPromise.then((body) => {
```
stefanpenner: rather then `done` lets use: `return bodyPromise.then` That way rejections become quick test failures, rather then timeouts.

nlfurniss: Yes, def good suggestion

addon/mixins/adapter-fetch.js
```diff
@@ -109,6 +109,23 @@ export function mungOptionsForFetch(_options, adapter) {
...
+ * Function checks that there is a body in the response. Otherwise JSON.parse will throw an error.
+ * If there is not, return an empty object and let Ember data handle the empty response
+ * @param {Response} response
+ * @returns {Promise}
+ */
+export function determineBodyPromise(response) {
+  let bodyPromise;
+  const contentLength = response.headers.get('content-length');
+
+  if (contentLength && Number(contentLength)) {
```
stefanpenner: sorry to nit-pick curious about the following possible values of `Number(contentLength)`

* `0` empty payload seems good
* '-1` `response.json()` may be unhappy

I suspect the if statement should be `if (contentLenNumber(contentLength) > 0) { ... }`

stefanpenner:
released as v3.0.1 🎉

PR #40, opened 2017-06-19
Better handle empty body responses

nlfurniss:
Instead of relying on the `Content-Length` header, we should assume that response's with status code of 204, 205, or `HEAD` requests won't have bodies and all other responses will.

@stefanpenner

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -110,21 +110,23 @@ export function mungOptionsForFetch(_options, adapter) {
   return options;
 }
 /**
- * Function checks that there is a body in the response. Otherwise JSON.parse will throw an error.
- * If there is not, return an empty object and let Ember data handle the empty response
+ * Function that determines what sort of Promise to return for the response's body.
+ * If the response has a status code of 204 (No Content) or 205 (Reset Content) or if the request method was 'HEAD',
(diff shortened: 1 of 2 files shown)


addon/mixins/adapter-fetch.js
```diff
@@ -110,21 +110,23 @@ export function mungOptionsForFetch(_options, adapter) {
...
   let bodyPromise;
+  const status = response.status;
 
-  // Content-Length is returned as a string or null, so we convert to a number.
-  const contentLength = Number(response.headers.get('content-length'));
-
-  if (contentLength > 0) {
-    bodyPromise = response.json();
+  if (status === 204 || status === 205 || requestData.method === 'HEAD') {
+    bodyPromise = RSVP.Promise.resolve({data: null});
```
nlfurniss: @runspired, is this the correct way to indicate to ED that the body is empty? It works, but I want to make sure it's correct.

stefanpenner: I believe so, but lets check with @runspired

runspired: Checking but I think this is problematic, ED treats explicit null separately from a missing attribute iirc.

runspired: @nlfurniss you're gtg, ED will do the right thing here.

stefanpenner:
released as v3.0.2 🎉

PR #42, opened 2017-06-21
Add hook into fetch behavior

nlfurniss:
@stefanpenner


stefanpenner:
@nlfurniss LGTM, one request is to add a quick test. Then we can :shipit:

nlfurniss:
Added tests, let me know if you think they look good enough

stefanpenner:
released as v3.1.0 🎉

PR #43, opened 2017-06-23
Fix logic for automatically adding a Content-Type header

nlfurniss:
Originally this logic just checked if the `Content-Type` header already existed, but it needs to check whether the request is a GET request and whether the request has a body.

Original logic from RestAdapter#ajaxOptions:

```
if (hash.data && type !== 'GET') {
  hash.contentType = 'application/json; charset=utf-8';
  hash.data = JSON.stringify(hash.data);
}
```

Removed default content-type header from test adapter to make tests easier to understand.


stefanpenner:
released as v3.1.1 🎉

PR #44, opened 2017-06-24
Rejection must be forwarded

cibernox:
@stefanpenner This contains your in #37 (plus a fixed typo) and a failing test that was failing before your change and it is working after.


stefanpenner:
released as v3.1.2 🎉

PR #46, opened 2017-06-26
Update README so that it works in FastBoot by default.

kratiahuja:
The README pointed to using the browser version of `fetch` by default. Make it point to use `ember-fetch/ajax` so that it works easily in fastboot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,13 @@ HTML5 [fetch](https://fetch.spec.whatwg.org) polyfill from [github](https://gith
 ## Usage
 
 ```js
-import fetch from 'fetch';
+import fetch from 'ember-fetch/ajax';
 import Ember from 'ember';
 
 export default Ember.Route.extend({
   model() {
     return fetch('/my-cool-end-point.json').then(function(response) {
-      return response.json();
+      return response;
     });
   }
 });


README.md
```diff
@@ -16,13 +16,13 @@ HTML5 [fetch](https://fetch.spec.whatwg.org) polyfill from [github](https://gith
...
 ```js
-import fetch from 'fetch';
+import fetch from 'ember-fetch/ajax';
 import Ember from 'ember';
 
 export default Ember.Route.extend({
   model() {
     return fetch('/my-cool-end-point.json').then(function(response) {
-      return response.json();
+      return response;
```
stefanpenner: using `ajax` the second `then` can be elided all-together.

stefanpenner: but I'm unsure why this needed to change at all for fastboot?

kratiahuja:
@stefanpenner sorry my bad. for a second i got confused between `import fetch from 'fetch';` and `import fetch from 'ember-fetch/ajax';` as two different things in fastboot. Going to close this.

README.md
tchak: I am not sure we should encourage the use of `ember-fetch/ajax`, unless we have a very good reason...

PR #49, opened 2017-06-29
Mimic $.ajax behavior, improve robustness

nlfurniss:
jQuery.ajax behaviors to mimic:
* Default to method `GET` if one is not provided.
* If the `url` in options already has query params, handle this when we process `options.data` to add additional query params.

Increase robustness:
* `headersToObject` should not error if no headers come back from the server.
* `ajaxError` should have a default  empty object for its `response` arg in the case where the fetch request throws an error and there is no response.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -71,7 +71,11 @@ function add(s, k, v) {
  */
 export function headersToObject(headers) {
   let headersObject = {};
-  headers.forEach((value, key) => headersObject[key] = value);
+
+  if (headers) {
+    headers.forEach((value, key) => headersObject[key] = value);
+  }
+
   return headersObject;
 }
 /**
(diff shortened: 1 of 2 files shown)


addon/mixins/adapter-fetch.js
```diff
@@ -90,12 +95,16 @@ export function mungOptionsForFetch(_options, adapter) {
...
+
+  // Default to 'GET' in case `type` is not passed in (mimicks jQuery.ajax).
+  options.method = options.type || 'GET';
 
   // GET and HEAD requests can't have a `body`
   if (options.data && Object.keys(options.data).length) {
     if (options.method === 'GET' || options.method === 'HEAD') {
-      options.url += `?${serialiazeQueryParams(options.data)}`;
+      // Test if there are already query params in the url (mimicks jQuey.ajax).
+      const queryParamDelimiter = options.url.match(QUESTION_MARK_REGEX) ? '&' : '?';
```
cibernox: Probably `indexOf` with a string is faster than `match` with a regexp

nlfurniss: 💯, updated

stefanpenner:
released as v3.2.6 🎉

PR #52, opened 2017-07-11
Fix and test support for URLSearchParams bodies

carlbennettnz:
github/fetch supports this, but it seems that the URLSearchParams global needs to be passed into the module.

I've added a test, but even with this fix it fails due to Phantom's lack of support for URLSearchParams. Not sure what you want to do here.

### DIFF
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -11,6 +11,9 @@
     if (global.Blob) {
       self.Blob = global.Blob;
     }
+    if (global.URLSearchParams) {
+      self.URLSearchParams = global.URLSearchParams;
+    }
 
     <%= moduleBody %>
 
--- a/tests/acceptance/root-test.js
+++ b/tests/acceptance/root-test.js
@@ -79,6 +79,30 @@ test('posting a form', function(assert) {
     assert.equal(data.name, 'World');
   });
 });
+test('posting a search params form', function(assert) {
+  server.post('/upload', function(req) {
+    assert.equal(req.requestHeaders['content-type'], 'application/x-www-form-urlencoded;charset=UTF-8');
+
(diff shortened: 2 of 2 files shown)


tchak:
Thank you for the contribution @carlbennettnz. I think it's good to give more options to how people will use `fetch`, but we also need to keep in mind the fact that`ember-fetch` is ment to work across browser and `fastboot`. It means if we going to support `URLSearchParams` we need to polyfill it in node. I have an addon with polyfills for `URL` and `URLSearchParams` https://github.com/tchak/ember-url. Feel free to use it if you need support for `URLSearchParams`.

carlbennettnz:
Ah, good point. I might be wrong here, but wouldn't that polyfill still not fix this problem? Detection of URLSearchParams is failing even if it exists.

tchak:
ah you mean the fact that in `fetch` polyfill `URLSearchParams` detection relies on the presence of it in the scope. Yeah. Not sure what to do about it...

carlbennettnz:
Yeah, exactly. This change wouldn't cause any errors in fastboot land if you're not trying to use URLSearchParams. Can we just require that you bring your own polyfill if you want to use URLSearchParams and Fastboot together?

tchak:
The problem is that your change will not enable the use of `URLSearchParams` in fastboot even with polyfill. It almost as if we should set `URLSearchParams` to `true` in our internal "global" env. Otherwise we are creating undefined behaviour.

carlbennettnz:
Oh yeah, I guess because the polyfill doesn't define a global as is expected by github/fetch, it still wouldn't work. Hmm...

stefanpenner:
This is part of 4.0.1 now, thanks for the PR. Sorry for letting it sit.

PR #57, opened 2017-07-22
Upgrade node-fetch

sivakumar-kailasam:
Node-fetch released a patch release which fixes a bug where the module `encoding` was removed as a dependency. This seems to be causing issues in fastboot deployments where`encoding` is missing. This fixes the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "broccoli-stew": "^1.4.2",
     "broccoli-templater": "^1.0.0",
     "ember-cli-babel": "^6.0.0",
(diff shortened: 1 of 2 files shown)


sivakumar-kailasam:
Please release a patch version on merging this PR

rwjblue:
AFAICT, this change seems fine but is not needed. The newer version is already allowed in the existing caret range. A `yarn upgrade` should grab the newer version just fine...

sivakumar-kailasam:
Whoops missed that!

Issue #58, opened 2017-07-31
Handling error payloads in the adapter mixin


webark:
In the [ajaxError method](https://github.com/ember-cli/ember-fetch/blob/master/addon/mixins/adapter-fetch.js#L244-L255) only the `response.statusText` is being passed to the `this.parseErrorResponse` method, but if a error payload has been passed back, it places the body in a `_bodyText` [property](https://github.com/github/fetch/blob/master/fetch.js#L388).

It would be nice to have a new method that is called which passes in the raw `response` that one could extend to pull out what is desired.

```javascript
  ajaxError(error, response = {}, requestData) {
    if (error instanceof Error) {
      return error;
    } else {
       return this.handleResponse(
        response.status,
         headersToObject(response.headers),
        this.parseErrorResponse(response.statusText) || error,
        requestData
      );
    }
  }
```

webark:
after doing some more investigation, and using a deployed server rather than a local or mocked data, it looks like the proper way to access the payload is a via promise from the `response.text` method. Going to do some investigation on the way to receive this with minimal invasiveness. Will update #59 when I find a solution.

PR #59, opened 2017-07-31
fix: add method to allow for overwriting the payload of an error response.

webark:
In relation to my earlier bug, #58, this allows a given adapter to overwrite the payload that gets passed back when a network error is encountered.


webark:
Closing in favor of the solution in #60.

PR #60, opened 2017-08-01
Brought the ajaxError and ajaxSuccess methods more inline with standard ember data methods

webark:
In ember data, both the ajaxError and ajaxSuccess methods receive the payload that was returned from the server. Bringing this adapter more inline with how those function.

Would be nice to be able to, in the future, rework these so that we can just call the super method and return those instead of rewriting all the logic ourselves in these methods.

Going to close #59 in favor of this solution.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -183,15 +183,15 @@ export default Ember.Mixin.create({
       })
       .then((response) => {
         return RSVP.hash({
-          responseData: determineBodyPromise(response, requestData),
-          response
+          response,
+          payload: determineBodyPromise(response, requestData)
         });
       })
-      .then(({response, responseData}) => {
+      .then(({ response, payload }) => {
         if (response.ok) {
-          return this.ajaxSuccess(this, response, responseData, requestData);
+          return this.ajaxSuccess(this, response, payload, requestData);
         } else {
-          throw this.ajaxError(this, response, requestData, responseData);
+          throw this.ajaxError(this, response, requestData, payload);
         }
       });
   },
@@ -260,11 +260,11 @@ export default Ember.Mixin.create({
     if (responseData instanceof Error) {
       return responseData;
     } else {
-      const parsedResponse = this.parseFetchResponseForError(response, responseData);
-      return this.handleResponse(
+      const parsedResponse = adapter.parseFetchResponseForError(response, responseData);
+      return adapter.handleResponse(
         response.status,
         headersToObject(response.headers),
-        this.parseErrorResponse(parsedResponse) || responseData,
+        adapter.parseErrorResponse(parsedResponse) || responseData,
         requestData
       );
     }
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -155,6 +155,9 @@ export function determineBodyPromise(response, requestData) {
     try {
       payload = JSON.parse(payload);
     } catch(error) {
+      if (!(error instanceof SyntaxError)) {
+        throw error;
+      }
       const status = response.status;
       if (response.ok && (status === 204 || status === 205 || requestData.method === 'HEAD')) {
         payload = { data: null };
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -143,9 +143,11 @@ export function mungOptionsForFetch(_options, adapter) {
 }
 /**
  * Function that always attempts to parse the response as json, and if an error is thrown,
- * returns an object with 'data' set to null if the response is a sucess, and
- * the plain payload if the response was an error.
+ * returns an object with 'data' set to null if the response is
+ * a sucess and has a status code of 204 (No Content) or 205 (Reset Content) or if the request method was 'HEAD',
+ * and the plain payload otherwise.
  * @param {Response} response
+ * @param {Object} requestData
  * @returns {Promise}
  */
 export function determineBodyPromise(response, requestData) {
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -179,14 +179,20 @@ export default Ember.Mixin.create({
 
     return this._ajaxRequest(hash)
       .catch((error, response, requestData) => {
-        throw this.ajaxError(error, response, requestData);
+        throw this.ajaxError(this, response, requestData, error);
       })
       .then((response) => {
+        return RSVP.hash({
+          responseData: determineBodyPromise(response, requestData),
+          response
+        });
+      })
+      .then(({response, responseData}) => {
         if (response.ok) {
-          const bodyPromise = determineBodyPromise(response, requestData);
-          return this.ajaxSuccess(response, bodyPromise, requestData);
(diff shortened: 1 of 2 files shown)


kratiahuja:
cc: @nlfurniss you might be interested in this.

webark:
@nlfurniss @stefanpenner was just wondering if there was anyone who could look into this!

addon/mixins/adapter-fetch.js
```diff
@@ -179,14 +179,20 @@ export default Ember.Mixin.create({
...
-        throw this.ajaxError(error, response, requestData);
+        throw this.ajaxError(this, response, requestData, error);
       })
       .then((response) => {
+        return RSVP.hash({
+          responseData: determineBodyPromise(response, requestData),
+          response
+        });
+      })
+      .then(({response, responseData}) => {
```
webark: I'll add a test for this, but that is the main thing i was wanting to change his n all this. Without waiting for the response of an error to be parsed, there is no way to requests with failing status codes that contain a meaningful response body.

addon/mixins/adapter-fetch.js
```diff
@@ -211,44 +217,54 @@ export default Ember.Mixin.create({
...
+      return responseData;
     } else {
-       return this.handleResponse(
+      const parsedResponse = this.parseFetchResponseForError(response, responseData);
+      return this.handleResponse(
         response.status,
-         headersToObject(response.headers),
-        this.parseErrorResponse(response.statusText) || error,
+        headersToObject(response.headers),
+        this.parseErrorResponse(parsedResponse) || responseData,
```
webark: It's whatever you choose to return from either the raw response or the payload. By default, the `adapter.parseFetchResponseForError(response, responseData);` will be returning the `responseData` which will be a json object.

webark:
To keep it to as close to where it was before, added back in the status and method checks for choosing when to return the `{ data: null }` response.

addon/mixins/adapter-fetch.js
nlfurniss: But since there is no guarantee that there is a response payload, it seems like `parseFetchResponseForError` should default to the `statusText` which we know will be there.

tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -348,6 +359,47 @@ test('determineBodyResponse returns an empty object when the request method is \
...
+  assert.expect(2);
+
+  this.errorAdapter = JSONAPIAdapter.extend(AdapterFetchMixin, {
+    _fetchRequest() {
+      const response = new Response(``, {status: 500});
+      return Ember.RSVP.Promise.resolve(response);
+    }
+  }).create();
+
+  const fetchReturn = this.errorAdapter.ajax({url: '/foo'});
```
nlfurniss: same as above

tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -348,6 +359,47 @@ test('determineBodyResponse returns an empty object when the request method is \
...
+      return Ember.RSVP.Promise.resolve(response);
+    },
+    parseFetchResponseForError() {
+      return {
+        errors: ['myOverWrittenError']
+      }
+    }
+  }).create();
+
+  const fetchReturn = this.errorAdapter.ajax({url: '/foo'});
```
nlfurniss: to remove confusion, can you change this to `ajax('/foo')` to match ajax's signature in this mixin

addon/mixins/adapter-fetch.js
```diff
@@ -211,44 +221,54 @@ export default Ember.Mixin.create({
...
+  parseFetchResponseForError(response, responseData) {
+    return responseData;
   },
 
   /**
-   * @param {Error} error
+   * @param {Object} adapter
    * @param {Object} response
    * @param {Object} requestData
+   * @param {Object} responseData
```
nlfurniss: should be`payload` throughout this method to be consistent.

addon/mixins/adapter-fetch.js
```diff
@@ -211,44 +221,54 @@ export default Ember.Mixin.create({
...
-      }
-    });
+
+/**
+ * Allows for the error to be selected from either the
+ * response object, or the response data.
+ * @param {Object} response
+ * @param {Object} responseData
+ */
+  parseFetchResponseForError(response, responseData) {
```
nlfurniss: should be`payload` throughout this method to be consistent

addon/mixins/adapter-fetch.js
nlfurniss: what about adding the error as another arg (use case line 186)? that way it's not confusing where `payload` can be json or an Error.

addon/mixins/adapter-fetch.js
webark: my only concern here about payload, is that ember data doesn't call it payload in the ajaxSuccess method https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1507 so i was wanting to use thier verbiage to maintain consistency. (even though they are inconsistent between thier error and success methods)

addon/mixins/adapter-fetch.js
webark: ya, I guess this changes a bit from ember data's method, https://github.com/emberjs/data/blob/171102c0106917fd4eae206f5e86e65b716d7713/addon/adapters/rest.js#L1507 where they are able to pull the response out of the `jqXHR` while due to how the response works with fetch, we are unable to do that directly. I'll think about this and get back to you.

webark:
@nlfurniss @stefanpenner any more thoughts on this?

tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -374,7 +374,7 @@ test('parseFetchResponseForError is able to be overwritten to mutate the error p
     }
   }).create();
 
-  const fetchReturn = this.errorAdapter.ajax({url: '/foo'});
+  const fetchReturn = this.errorAdapter.ajax({url: '/trigger-a-server-error-with-content'});
```
webark: I'm not exactly sure what you are meaning by this.

nlfurniss: You're passing in an object to `ajax`, but it should be `ajax('/trigger-a-server-error-with-content')`

webark: ohh.. sure.

addon/mixins/adapter-fetch.js
```diff
@@ -141,25 +142,28 @@ export function mungOptionsForFetch(_options, adapter) {
...
-  if (status === 204 || status === 205 || requestData.method === 'HEAD') {
-    bodyPromise = RSVP.Promise.resolve({data: null});
-  } else {
-    bodyPromise = response.json();
-  }
-  return bodyPromise;
+  return response.text().then(function(payload) {
+    try {
+      payload = JSON.parse(payload);
+    } catch(error) {
```
webark: that's a good idea. Done.

webark:
@nlfurniss let me know if you have any other ideas or suggestions!

webark:
thanks @nlfurniss !! i'll keep my eyes out for any issues that might arise.

Issue #45, opened 2017-06-25
Avoid import when targets support native `fetch`


rwjblue:
We can use `this.project.targets` to determine if the `whatg-fetch` polyfill is required for the browser builds, and avoid `app.import`ing it when it is not needed.

See [browser support](http://caniuse.com/#search=fetch).  Roughly includes all evergreen browsers (FireFox, Edge, Chrome, Safari)...

cibernox:
I agree, but there was some fear for two things:

- Possible edge cases in which the polyfill doesn't behave perfectly
- The problem that pretender.js will not work with fetch, and it is VERY popular. We need pretender-fetch to happen.

rwjblue:
> Possible edge cases in which the polyfill doesn't behave perfectly

I'm not sure what you mean here. 

> The problem that pretender.js will not work with fetch, and it is VERY popular. We need pretender-fetch to happen.

I believe that _someone_ was working on this already, do you recall that?

cibernox:
> I'm not sure what you mean here.

I mean that @stefanpenner was concerned that perhaps people in development had modern targets and things work but when they deploy to prod, with the polyfill, something is different in some subtle way. I don't buy that argument a lot tho.

> I believe that someone was working on this already, do you recall that?

I recall, it was me 🤣 . I really need to come back to that.

kratiahuja:
I am curious to understand why fetch will not work with pretender? Pretender just replaces the XHR layer with Fake XHR AFAIU. If fetch (with the polyfill) is calling XHR under the hood, you don't need a pretender-fetch. Example: if you use jQuery.ajax pretender doesn't do anything special.

rwjblue:
>  If fetch (with the polyfill) is calling XHR under the hood, you don't need a pretender-fetch. 

The goal of this issue is to **avoid** including the polyfill at all, when the browsers that the application supports already have native `fetch`.

kratiahuja:
Agreed, my question is more around to understand why pretender will not work with `fetch` in testing (if your tests are already running against browsers that support fetch)? pretender should be agnostic to the high-level network interface being used (whether fetch/jQuery.ajax) etc.

rwjblue:
Pretender currently only captures requests that go through XHR. When native fetch is used its requests will no longer funnel through XHR and therefore Pretender will not be able to intercept. @cibernox has a spike for a "fake fetch" (akin to FakeXHR which pretender uses now), which should address this concern (though I actually do not consider it a blocker).

[...]

PR #63, opened 2017-08-14
use browser fetch if exists

xg-wang:
[GitHub/fetch](https://github.com/github/fetch/blob/master/fetch.js#L4) expects a `window` that contains a fetch, Headers, Response... as the passed in `self` so it can skip the unnecessary polyfill. But in ember-fetch the passed in `self` is actually the exported object, which is `{}` at first.


stefanpenner:
This project explicitly forces the polyfil in all cases. This is because the polyfill is not complete, and switching between modes automatically is dangerous.

I would be open to a config setting to allow this as an opt-in.

xg-wang:
@stefanpenner Just added the option which could be set in 'ember-cli-build.js' as
```
  var app = new EmberAddon(defaults, {
    'ember-fetch': {
      forcePolyfill: false
    }
  });
```

rwjblue:
https://github.com/ember-cli/ember-fetch/issues/45 is also somewhat related.

assets/browser-fetch.js.t
```diff
@@ -2,15 +2,30 @@
...
+
+    var supportProps = [
+      'FormData',
+      'FileReader',
+      'Blob',
+      'URLSearchParams',
+      'Symbol',
+      'iterator',
+      'ArrayBuffer'
+    ];
```
tchak: I think if we are to add all of these we should also add `fastboot` polyfills in order to ensure parity.

xg-wang: Why are we adding `FormData`, `FileReader`, `Blob` to self? They are also not added to Fastboot fetch.

xg-wang:
To ensure parity with Fastboot polyfills, supported properties(FormData, FileReader, Blob, etc) should only be used internally in github/fetch but not exported.
We previously export FormData, FileReader, Blob but they were actually missing in Fastboot.

stefanpenner:
We explicitly don't do this right now,

xg-wang:
@stefanpenner 
I still think we should use native fetch if possible. It is becoming stable and we need to have it available for Ember developers. This will also make other projects' support for `fetch` easier. There are people who need the native `fetch` and won't have those edge cases on polyfilled browsers. Putting this new feature behind a flag as implemented in this PR is a good choice for now.

The Github polyfill for `fetch` only take effect if `window.fetch` doesn't exist, IMO we should make `ember-fetch` behave same to it, and as for now we need to explicitly point out this difference in the readme (that `ember-fetch` never uses native fetch).

Make it more clear, my preferred roadmap for `ember-fetch` is
1. Explicitly point out this project discard native fetch and always use XHR next to the reference to Github polyfill
2. Finish this PR to provide a feature flag which needs to be explicitly turned on to provide access to native fetch - on par with the GitHub polyfill. When doing this we also need to check for testing so that `pretender` don't fail - and have this caveat highlighted in the readme to let people be aware.
3. Wait until https://github.com/pretenderjs/pretender/issues/60 close and correlated changes in mirage land, then drop the testing check mentioned above - this will ensure same behaviour in prod and test
4. Default access to native fetch for a major version bump

We need to discuss the potential disparity between fastboot fetch and browser fetch, and where should changes be made.

@rwjblue I'm not sure about #45. We don't need to avoid import when targets have `fetch` since Github polyfill uses native fetch if exists.

stefanpenner:
@xg-wang this sounds mostly reasonable. One key area that will need to still be dealt with is that IE11 (a supported platform) does not have fetch.

So for users that fall back to the polyfil, what features will they loose, and for those lost features what can we do to mitigate that pain.

It's never fun to realize that some important aspect of your app is non-functional once you test in IE11.

tchak:
@stefanpenner as far as I can tell, the main missing feature from polyfil is streams. I have not investigated it yet, but is it really unreasonable to polyfil by using progress events on XHR?

xg-wang:
@stefanpenner Github polyfill
> is a polyfill that implements a subset of the standard Fetch specification

I will figure out the supplement set and see if we can do a warning message for those.

For native `fetch` it should be fine to leave it to the browser vendor behavior.

stefanpenner:
@tchak / @xg-wang thoughts on simply wrapping native fetch, in something that hides stream stuff by default? In a way that errors with a helpful message (and opt-opt instructions). Basically just catching the cases where the polyfill will fall short.

^^ would be for if we can't reasonably polyfil the remaining features.

xg-wang:
We have these choices:
1. Forbid experimental features that're only available in subset of targets and throw message
2. Allow access to experimantal features and throw warnings, and offer `ember-fetch` APIs for checking availabiliry. (This is something like `this.get('fastboot.isFastBoot')`)
3. Provide polyfill for experimental features based on option 2
4. Expose native `fetch` just as Github polyfill and do nothing

My thought: 
- These are also different levels of autonomy that we give to ember developers, and IMO should also take the browser support situation into consideration. 
- Experimental features like `Streams` and `Priority hints` (need a relatively more complete list though it may update frequently) are cool and people may want to try them out for available browsers.

My opinion: 
- I like option 2 but it takes some time to implement & figure out the complete list. 
- It would be reasonable and safe to start with option 1, and provide warnings for some features that we know people might be using "mistakenly". 
- I don't want to introduce more polyfills, would rather wait for native support.

@stefanpenner

tchak:
The problem with leaving streams out is that it is the only way with `fetch` to show progress for uploads. See #99. Can we try to use https://github.com/creatorrr/web-streams-polyfill?

I feel that if we want for `fetch` to be a serious alternative to `xhr` it needs some level of feature parity.

stefanpenner:
> We have these choices:

We also have, in development/test only use polyfil, and in prod use real. 

> The problem with leaving streams out is that it is the only way with fetch to show progress for uploads. See #99. Can we try to use https://github.com/creatorrr/web-streams-polyfills

For users who want to use streams, we can allow a "alwaysUseNativeFetch" style option, which opts out of the above safety.

Thoughts?

stefanpenner:
> can we try to use https://github.com/creatorrr/web-streams-polyfills

I would love to here how well this works (tell me the downsides), and the impact of it.

assets/browser-fetch.js.t
xg-wang: Just walked through PR list, these have been merged and are exported, but not in Fastboot...
#108 #93 #76

xg-wang:
I updated this patch so we can have some real code stuff to talk about.

index.js
```diff
@@ -75,15 +75,16 @@ module.exports = {
       } while (current.parent.parent && (current = current.parent));
     }
 
-    this.buildConfig = target.options['ember-fetch'] || {forcePolyfill: true};
+    this.buildConfig = target.options['ember-fetch'] || { preferNative: false };
```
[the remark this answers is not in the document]
stefanpenner: This sounds good. To help forks deal with the upgrade, could we detect if they use `forcePolyfill` property, and throw a useful error message?

Once released, this will also be a major bump.

cibernox:
While not technically related, one problem with using native fetch right now is that pretender and ember-cli-mirage won't work. It's not per-set a problem of this library, but will bite people if we go ahead and do this. Someone needs to make pretender work with fetch.

stefanpenner:
@cibernox making pretender work with native fetch sounds good. I believe @xg-wang's proposal is to run the polyfill fetch during dev/tests, but native in production. This would allow pretender (used in dev) to "just work".

We should obviously follow up with a fetch solution for Pretender (https://github.com/pretenderjs/pretender/issues/60). This shouldn't take much time, and I believe @xg-wang will be working on fetch + pretender compat next.

index.js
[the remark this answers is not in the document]
xg-wang: I will squash the commits once I have a more meaningful implementation, right now it's more of a brainstorm

xg-wang:
For this change, open the dummy app and see console to confirm it's working:

- `ember s`: XHR finished loading: GET "http://localhost:4200/slow-data.json".
- `ember b --prod`: Fetch finished loading: GET "http://127.0.0.1:8080/slow-data.json".
- `ember test`: Pass

Something warn users like this:
```
 ❯ ember b --prod                                                                                          [22:33:24]
[ember-fetch] Prefer native fetch enabled, please be aware of browser and pretender support.
```

index.js
```diff
@@ -63,16 +63,29 @@ module.exports = {
...
-       target = current.app || app;
-      } while (current.parent.parent && (current = current.parent));
+    // We access app config and pass in the environment: 'development', 'test' or 'production'
+    // The current idea is we let user to decide situation to prefer native fetch
+    this.fetchConfig = app.project.config(app.env)['ember-fetch'] || {
+      preferNative: false
+    };
+    if (this.fetchConfig.preferNative === true) {
+      this.ui.writeInfoLine(
+        '[ember-fetch] Prefer native fetch enabled, please be aware of browser and pretender support.'
```
stefanpenner: I'm unsure about this logging. If we do think this logging is important, it is too vague, rather I would recommend something like:

```
[ember-fetch] fetch polyfil disabled, see: https://github.com/ember-cli/ember-fetch#something-more-details-about-this for more details
```

---

That being said, do you feel that seeing this log everytime you boot up the app is valuable? We don't do this for other polyfils, what makes this special.

assets/browser-fetch.js.t
```diff
@@ -38,7 +45,7 @@
       self['default'] = function() {
         pending++;
 
-        return self.fetch.apply(self, arguments).then(function(response){
+        return self.fetch.apply(preferNative ? global : self, arguments).then(function(response){
```
stefanpenner: I believe this is pulling from the wrong config, see https://github.com/ember-cli/ember-fetch/pull/63/files#r190637528 for more details.

tests/dummy/config/environment.js
```diff
@@ -45,6 +45,9 @@ module.exports = function(environment) {
 
   if (environment === 'production') {
     // here you can enable a production-specific feature
+    ENV['ember-fetch'] = {
+      preferNative: true
```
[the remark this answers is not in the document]
xg-wang: The intention was to let the user control the behavior under different environment, once pretender lands the fetch support ember-fetch no longer needs to consider different environment config. I'm going to move it back to the build config.
@stefanpenner

ember-cli-build.js
```diff
@@ -5,6 +5,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function(defaults) {
   let app = new EmberAddon(defaults, {
     // Add options here
+    'ember-fetch': {
+      preferNative: true
```
xg-wang: hmm, maybe should be false here for the dummy app?

package.json
```diff
@@ -16,20 +16,23 @@
...
     "abortcontroller-polyfill": "^1.1.9",
     "broccoli-concat": "^3.2.2",
     "broccoli-merge-trees": "^2.0.0",
     "broccoli-stew": "^1.4.2",
-    "broccoli-templater": "^1.0.0",
+    "broccoli-templater": "^2.0.1",
     "ember-cli-babel": "^6.8.2",
     "node-fetch": "^2.0.0-alpha.9",
-    "yetch": "^0.0.1"
+    "whatwg-fetch": "https://github.com/github/fetch"
```
xg-wang: TODO, change this when new version released

tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -15,512 +12,615 @@ import AdapterFetchMixin, {
...
+    assert.expect(1);
 
-  const newText = 'intercepted';
-  this.JSONAPIAdapter._fetchRequest = () => newText;
-  const fetchReturn = this.JSONAPIAdapter._ajaxRequest({});
-  assert.equal(fetchReturn, newText);
-});
+    const fetchReturn = this.JSONAPIAdapter._ajaxRequest({});
+    // if in pretender with native fetch testing, pretender will polyfill native fetch
+    assert.ok(fetchReturn instanceof Promise || fetchReturn instanceof window.Promise);
```
xg-wang: Highlight
This is the only change other than formatting. We have RSVP.Promise but pretender will have the Promise provided specified in `ember-cli-pretender`

xg-wang:
Review needed @tchak @stefanpenner @rwjblue, should we mention stream and other fetch polyfill caveats?

xg-wang:
This PR is ready, need some review. @rwjblue @stefanpenner @tchak

index.js
```diff
@@ -112,20 +123,17 @@ module.exports = {
...
 // that can be used by the Ember app.
 function treeForBrowserFetch() {
-  var fetchPath = require.resolve('yetch/polyfill');
-  var abortcontrollerPath = require.resolve('abortcontroller-polyfill');
-
-  var expandedFetchPath = expand(fetchPath, 'dist/yetch-polyfill.js');
-  var expandedAbortcontrollerPath = expand(abortcontrollerPath, 'abortcontroller-polyfill-only.js');
-
-  var polyfillTree = concat(new MergeTrees([find(expandedFetchPath), find(expandedAbortcontrollerPath)]), {
+  const fetchTree = path.dirname(require.resolve('@xg-wang/whatwg-fetch'));
```
stefanpenner: lets leave a comment above this line, briefly explaining why it is a fork and when we hope to get back onto the mainline.

test/prefer-native-test.js
```diff
@@ -0,0 +1,95 @@
...
+  _fetch.default();
+});
+require('fetch-test');
+`;
+
+[true, false].forEach(flag => {
+  describe(`Build browser assets with preferNative = ${flag}`, function() {
+    let output, subject, addon;
+
+    beforeEach(async function() {
```
stefanpenner: we can drop `async` here, as it doesn't appear we have an await in the function

index.js
rwjblue: Maybe we can actually have an issue on the fork with this info, and have the comment here cross link that?

xg-wang: Open 2 issues where the fork is being used, I will open another one under ember-fetch and cross link them.

https://github.com/pretenderjs/pretender/issues/232
https://github.com/rwjblue/ember-cli-pretender/issues/63

test/prefer-native-test.js
xg-wang: good catch

xg-wang:
yarn registry seems not stable these days

CI passed on mine repo
https://travis-ci.com/xg-wang/ember-fetch/builds/79599780

stefanpenner:
released as v5.1.0 🎉🎉🎉🎉🎉🎉

PR #64, opened 2017-08-18
Bump Ember-CLI, ember-cli-babel

mixonic:
In order to run apps without a dependency on ember-cli-shims, addons that reference the "ember" module must be running ember-cli-babel 6.8.0+. Bump that dependency and generally update Ember-CLI. Expand ember-try to the 2.12 LTS.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -50,6 +50,14 @@ module.exports = {
       }
     },
     {
+      name: 'ember-lts-2.12',
+      npm: {
+        devDependencies: {
+          'ember-source': '~2.12.0'
+        }
+      }
+    },
+    {
       name: 'ember-release',
       bower: {
         dependencies: {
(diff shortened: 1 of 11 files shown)


config/ember-try.js
```diff
@@ -50,6 +50,22 @@ module.exports = {
       }
     },
     {
+      name: 'ember-lts-2.12',
+      bower: {
+        dependencies: {
+          'ember': 'components/ember#lts-2-12'
```
rwjblue: Can't pull lts-2-12 from bower (the [blueprints do this](https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/addon-config/ember-try.js#L20-L27)).

mixonic:
Pushed. I've also been testing this more, and it alone is not enough to drop ember-cli-shims. The file [`browser-fetch.js`](https://github.com/ember-cli/ember-fetch/blob/master/assets/browser-fetch.js.t#L2) has a runtime dependency. I'll follow up on that in another PR.

PR #65, opened 2017-08-19
Drop runtime Ember module usage

mixonic:
To permit this library to be used when there is no ember-cli-shims addon in the host app, drop the modules usage in `browser-fetch.js`. Instead, reach for the global `Ember` object directly.

### DIFF
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -1,7 +1,7 @@
 (function (global) {
-  define('fetch', ['ember', 'exports'], function(Ember, self) {
+  define('fetch', ['exports'], function(self) {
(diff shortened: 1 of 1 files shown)


rwjblue:
I'm concerned by this approach. Replacing the imports means this library _requires_ global Ember. In most cases we are trying to push folks towards the more explicit imports that _happen to be_ polyfilled into global Ember usage.

I'd rather see us move this to actually being transpiled, so that we can leverage the same thing that applications can...

rwjblue:
Meh, past me was not practical enough, though I still agree with my last statement it just isn't an issue for the foreseeable future. Sorry @mixonic.

PR #66, opened 2017-08-23
Don't process empty options.data, don't filter out 'null' query param values

nlfurniss:
Previously ([eabb092](https://github.com/ember-cli/ember-fetch/commit/eabb092cdc038a408f7733bc6d02b7bd08d0f4ea)), empty `options.data` objects weren't processed. This commit restores that logic.

Additionally, we were incorrectly removing queryParams with `null` values, when it should have only been params with a value of `undefined`. This has been corrected.

### DIFF
--- a/tests/unit/mixins/adapter-fetch-test.js
+++ b/tests/unit/mixins/adapter-fetch-test.js
@@ -209,7 +209,7 @@ test('mungOptionsForFetch adds string query params to the url correctly', functi
   assert.equal(options.url, `${baseUrl}?fastboot=true&a=1&b=2`, 'url that started with query params has more query params');
 });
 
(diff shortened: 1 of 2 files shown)


tests/unit/mixins/adapter-fetch-test.js
```diff
@@ -246,6 +246,28 @@ test('mungOptionsForFetch sets the request body correctly when the method is POS
   assert.deepEqual(options.body, stringifiedData);
 });
 
+test('mungOptionsForFetch does not process an empty \'data\' object', function(assert) {
```
stefanpenner: mung or munge  ?

nlfurniss: https://en.wikipedia.org/wiki/Mung_(computer_term)

Guess it can be spelled both ways? I've always had `mung` in this repo.

PR #75, opened 2017-10-03
Convert request body to JSON if it is a string

jcao219:
Hi, close this PR if it is not intended behavior.

In version 3.3.1, ember-fetch would convert string POST data to JSON, which is a behavior that we relied on.

Now in version 3.4.0, ember-fetch just leaves strings as is. So `foo` remains as `foo` when sent to the server. But, in 3.3.1 ember-fetch would convert `foo` to `"foo"`, a valid JSON serialized string, which is what the backend expects. I wrote a unit test to highlight this behavior.

I noticed that there's another unit test which seems to indicate that 3.4.0 breaks the old behavior intentionally, therefore I'll have to stringify all data before making the ajax call. I wonder about the design decision behind this, as it's strange how JSON.stringify gets called for objects, but not strings. I don't think this new design is bad, though, but I wish that the CHANGELOG made it clearer.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -116,14 +116,9 @@ export function mungOptionsForFetch(_options, adapter) {
       const queryParamDelimiter = options.url.indexOf('?') > -1 ? '&' : '?';
       options.url += `${queryParamDelimiter}${serialiazeQueryParams(options.data)}`;
     } else {
-      // NOTE: a request's body cannot be an object, so we stringify it if it is.
-      if (typeof options.data === 'string') {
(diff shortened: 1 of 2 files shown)


jcao219:
To clarify, this breaking change was caused by commit 691d1ba98099332813a354ff148082462b0bd84c . The overridden method normally stringifies string data. See: https://github.com/emberjs/data/blob/v2.14.10/addon/adapters/rest.js#L1120

nlfurniss:
Thanks @jcao219 for bringing this up. I've fixed the issue in #84 outside this PR since more tests needed to be added and other failing tests needed fixing.

If you see any other issues, please do open an issue or PR.

PR #76, opened 2017-10-09
add FormData support to fastboot

tchak:
We do expose `FormData` in the browser so it make sens to me to expose it also in fastboot.


kratiahuja:
Can you add tests?

tchak:
@kratiahuja right now there is no tests for fastboot in `ember-fetch` :( I agree we should have some but it seems like a completely separate (and not trivial issue).

stefanpenner:
@tchak anything we can do to close this one out?

PR #78, opened 2017-10-09
update node in appveyor (remove node 0.12)

tchak:
I don't really know what I am doing here :) I just realised that bunch of tests used to pass with 0.12...


tchak:
This is not fixing appveyor, but it may still be with it to update node versions

PR #83, opened 2017-10-11
remove app re-export of mixin

kellyselden:
I think this was added by mistake, and since the readme doesn't make use of it, I don't think it is a breaking change to remove it.


tchak:
I feel there is enough approval here to merge ;)

PR #89, opened 2017-10-20
Add config option to disable export default fetch module using node-f…

tsubomii:
1. Create configuration option to disable defining fetch using node-fetch in fastboot. 
2. Add tests to ensure build output is correct for fastboot environment.
3. Updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,15 @@ export default DS.RESTAdapter.extend(AdapterFetch, {
 });
 ```
 
+### Configuration
+This addon provides `fetch` in fastboot. The default `fetch` module is node-fetch, which uses node http or https underneath. If your app require different communication protocal on server side. You can disable using node-fetch by setting following config value in ember-cli-build.js to be true. Without setting this value means using node-fetch for server side `fetch`.
+```js
+let app = new EmberAddon(defaults, {
+    'ember-fetch': {
+      disableDefaultNodeFetch: true
+    }
+```
+
 further docs: https://github.com/github/fetch
 
 ### Browser Support
--- a/index.js
+++ b/index.js
@@ -38,26 +38,13 @@ module.exports = {
    */
   included: function(app) {
     this._super.included.apply(this, arguments);
-    
+
     let target = app;
 
     if (typeof this.import === 'function') {
-      target = this;
+        target = this;
     } else {
-      // If the addon has the _findHost() method (in ember-cli >= 2.7.0), we'll just
-      // use that.
-      if (typeof this._findHost === 'function') {
-        target = this._findHost();
-      }
-
-      // Otherwise, we'll use this implementation borrowed from the _findHost()
-      // method in ember-cli.
-      // Keep iterating upward until we don't have a grandparent.
-      // Has to do this grandparent check because at some point we hit the project.
-      let current = this;
-      do {
-       target = current.app || app;
-      } while (current.parent.parent && (current = current.parent));
+      target = this._findHostApp();
     }
 
     target.import('vendor/ember-fetch.js', {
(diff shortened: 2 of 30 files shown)


index.js
```diff
@@ -87,10 +74,46 @@ module.exports = {
...
+    return target;
+  },
+
+  //convenient method to determine if we should import the default node fetch module
+  //or skip. Let consumer app to import node fetch from elsewhere
+  _shouldImportDefaultNodeModule: function() {
+    var app = this._findHostApp();
+    var appOptions = (app && app.options) || {};
+    //make it backwards compatiable for apps doesn't set this config
+    if (typeof appOptions['ember-fetch'] === 'undefined') {
```
rwjblue: I'd suggest changing this conditional around a bit.  As it stands, setting `'ember-fetch'` to `{}` in `ember-cli-build.js` will disable the default behavior, but I don't think that is what we want.  I believe we would want to only disable if the value is explicitly `true`.

index.js
```diff
@@ -87,10 +74,46 @@ module.exports = {
...
+    } while (current.parent.parent && (current = current.parent));
+    return target;
+  },
+
+  //convenient method to determine if we should import the default node fetch module
+  //or skip. Let consumer app to import node fetch from elsewhere
+  _shouldImportDefaultNodeModule: function() {
+    var app = this._findHostApp();
+    var appOptions = (app && app.options) || {};
+    //make it backwards compatiable for apps doesn't set this config
```
rwjblue: s/compatiable/compatible/

index.js
```diff
@@ -38,26 +38,13 @@ module.exports = {
    */
   included: function(app) {
     this._super.included.apply(this, arguments);
-    
+
     let target = app;
 
     if (typeof this.import === 'function') {
-      target = this;
+        target = this;
```
rwjblue: indentation seems wrong here

index.js
```diff
@@ -141,7 +164,7 @@ function treeForBrowserFetch() {
 // we just happen to know that the passed tree only contains a single file.
 function normalizeFileName(tree) {
   return rename(tree, function() {
-    return 'ember-fetch.js';
+      return 'ember-fetch.js';
```
rwjblue: Indentation seems wrong here

index.js
```diff
@@ -150,4 +173,4 @@ function expand(input) {
   var file = path.basename(input);
 
   return dirname + '/{' + file + '}';
-}
+}
```
rwjblue: We should preserve newline at end of file

package.json
```diff
@@ -15,18 +15,26 @@
   "scripts": {
     "build": "ember build",
     "start": "ember server",
-    "test": "ember try:each"
+    "test": "ember try:each && mocha",
```
rwjblue: Can we split this into separate test commands?

```json
"scripts": {
  "test": "npm run test:browser && npm run test:node",
  "test:browser": "ember try:each",
  "test:node": "mocha"
}
```

package.json
```diff
@@ -15,18 +15,26 @@
...
+    "test:precook": "node node_modules/ember-cli-addon-tests/scripts/precook-node-modules.js"
   },
   "dependencies": {
+    "bower": "^1.8.2",
```
rwjblue: Please remove this, this project does not depend on `bower`

package.json
```diff
@@ -15,18 +15,26 @@
...
     "broccoli-funnel": "^1.2.0",
     "broccoli-stew": "^1.4.2",
     "broccoli-templater": "^1.0.0",
+    "ember-cli-addon-tests": "0.9.2",
```
rwjblue: This should be a devDependency

index.js
tsubomii: This utility function determines if we should define fetch module from node-fetch. It's reverse from the configuration flag disableDefaultNodeFetch. If _shouldImportDefaultNodeModule return true, node-fetch is define. If return false, it's not defined.

The logic is a bit flipped here because I want to make it backwards compatible. How should I make it more clear?

rwjblue: What I'm saying is that the logic is not quite right, you get node-fetch disabled in both of the following cases:

1. you set `'ember-fetch': { disableDefaultNodeFetch: true }`
2. you have `'ember-fetch': {}`

index.js
tsubomii: Fixed

README.md
```diff
@@ -43,6 +43,15 @@ export default DS.RESTAdapter.extend(AdapterFetch, {
 });
 ```
 
+### Configuration
+This addon provides `fetch` in fastboot. The default `fetch` module is node-fetch, which uses node http or https underneath. If your app require different communication protocal on server side. You can disable using node-fetch by setting following config value in ember-cli-build.js to be true. Without setting this value means using node-fetch for server side `fetch`.
+```js
+let app = new EmberAddon(defaults, {
+    'ember-fetch': {
+      disableDefaultNodeFetch: true
```
rwjblue: Any objection to changing this to `disableNodeFetch` (I don't think we need "default" in there)

index.js
tsubomii: Fixed :)

index.js
tsubomii: Fixed

index.js
tsubomii: Fixed

package.json
tsubomii: Great idea to separate them. The node test takes time, because if needs to install the fixture apps 👍

index.js
```diff
@@ -87,10 +74,48 @@ module.exports = {
...
+
+  //convenient method to determine if we should import the default node fetch module
+  //or skip. Let consumer app to import node fetch from elsewhere
+  //return true means we will define fetch from node-fetch
+  //return false means fetch in node will not be defined, it's user's responsiblity to define their own fetch in fastboot
+  _shouldImportDefaultNodeModule: function() {
+    var app = this._findHostApp();
+    var appOptions = (app && app.options) || {};
+    //make it backwards compatible for apps doesn't set this config
+    if (typeof appOptions['ember-fetch'] === 'undefined') {
```
rwjblue: Perhaps modifying to:

```js
let shouldDisableNodeFetch = appOptions['ember-fetch'] && appOptions['ember-fetch'].disableDefaultFetch === true;

return !shouldDisableNodeFetch;
```

^ this address my concern about `'ember-fetch': {}`

tsubomii: I see. I missed understood what you said earlier. Thanks!

stefanpenner:
@tsubomii / @rwjblue anything we can do to close this one out?

stefanpenner:
@xg-wang does your work supersede this, or ...

xg-wang:
@stefanpenner I think they are orthogonal. node-fetch doesn't touch browser.

PR #90, opened 2017-10-25
add fetch service

luxferresum:
The main reason for me to use `ember-ajax` in favor of `ember-fetch` is the service. I'm not 100% sure if this should be part of `ember-fetch` itself or in a seperate addon. However since `adapter-fetch` is located here as well, and it would be nice for the adapter mixin to use the ajax/fetch service as well, I think a fetch service should be provided by this addon.


rwjblue:
Can you help me understand the benefit that the service provides over the "simple" import?

rwjblue:
Is it about shared header handling?

luxferresum:
> Is it about shared header handling?
Yes! Thats one use-case. Or more generally: shared information.
Some Backends also use tokens in the URI for authentication.

Also it could be useful for logging or to provide a global AJAX loading indicator.

tchak:
@luxferresum another interesting approach is to use https://github.com/tchak/ember-wretch. It allows you to "curry" fetch options and by doing so share configuration.

nlfurniss:
You can already share headers across API calls by setting the `headers` prop on your application adapter, or any other adapter.

[ED docs](https://emberjs.com/api/ember-data/2.16/classes/DS.RESTAdapter)
[Usage in ember-fetch/mixin](https://github.com/ember-cli/ember-fetch/blob/master/addon/mixins/adapter-fetch.js#L100)

rwjblue:
@nlfurniss - sure, but that’s only for requests through ember-data’s store. This library services both ember-data users and folks that “just” want to fetch something directly...

nlfurniss:
Right, just letting people know that that’s half supported 😋

luxferresum:
@tchak interesting! However doesnt solve the problem because you still would have to create the service and put the global shared version there. This is about the same effort you need to implement a custom `fetch` service. Also you have the porblem, that probably you want to share theese headers between ember-data and all other ajax calls.

And we had all theese arguments already with `ember-ajax`! The only reason why `ember-ajax` exists is for the `ajax` service. So if this addon wants to replace `ember-ajax` I think the `fetch` service is required.

tchak:
I am not opposed to the idea of a service (it should contain large part of what is in the `ember-data` mixing right now I think) but I also wonder if mixin and service really belong in ember-fetch and shouldn't be a separate addon. I can see scenarios (particularly on mobile) when I want just `fetch` and right now I pay the payload price for the mixin.

luxferresum:
@tchak I agree. First I wanted to build a separate addon, however I think that mixing and the service should live on the same place, so I opened this PR.

I agree that payload size should be considered. However we already have the 7.89KB of the mixing, and I think the additional 614 Bytes of the fetch service shouldn't be the blocker for now.

In long term it could be a good idea to separate mixing & service from the pure `fetch` polyfill. The payload burden argument does go both ways. When dropping IE11 support in an app, I don't need a fetch polyfill at all.

Maybe it could be a way to split this addon into something like `ember-fetch-polyfill` and `ember-fetch-tools` and let `ember-fetch` install both?

Or we could just make it configurable. `ember-composable-helpers` does this. So we could opt-out of the service and the mixing with a configuration. And for the choice to include the polyfill we could inspect the `config/targets.js`.

webark:
I’m down for two addon. The adapter always seemed weird to be included in this one. 
The inclusion of the polyfill reading off the targets should happen regardless of say.. Adding in a service and addon based off a configuration seems weird.

stefanpenner:
I don't feel we should ship this as part of this add-on. Seems heavy weight to merely share headers.

webark:
Was a general “fetch service addon” ever implemented as a standard addon? Been using ember ajax for shared headers, retry logic, standard error handling, etc, and the like, but wanted to switch to fetch but not “reinvent the wheel” if one was already there.

PR #93, opened 2017-12-15
fix ArrayBuffer support

mohanyin:
Currently, this library doesn't let you send `ArrayBuffer`s using fetch, even though the underlying polyfill supports it.


mohanyin:
Hm... looks like PhantomJS doesn't support ArrayBuffer, which makes this pretty difficult to test. Does anyone have thoughts on how to proceed?

tests/acceptance/root-test.js
```diff
@@ -79,6 +79,30 @@ test('posting a form', function(assert) {
     assert.equal(data.name, 'World');
   });
 });
+test('posting an array buffer', function(assert) {
```
rwjblue: Wrap this test in an `if (typeof ArrayBuffer !== ‘undefined’) {`, which will get the tests passing on phantom (by not running this one) and on chrome...

rwjblue:
I think CI was just fixed in #95, can you rebase (so we can confirm AppVeyor is passing properly)?

mohanyin:
@rwjblue mind taking another look?

PR #96, opened 2017-12-30
[TS] ember-fetch returns an RSVP.Promise

mike-north:


mike-north:
@rwjblue mainly for the presence of `.finally()`, [while we wait for official adoption in ES6 promises](https://github.com/tc39/proposal-promise-finally).

stefanpenner:
Ah this has been around for awhile now. Is this still relevant, now that we have `finally` on promises? If so, let me know, I'l merge/release.

PR #114, opened 2018-06-05
Update broccoli-templater to 2.0.1.

rwjblue:
Fixes the following warnings on ember-cli@3.1:

```
[API] Warning: The .read and .rebuild APIs will stop working in the next Broccoli version
[API] Warning: Use broccoli-plugin instead: https://github.com/broccolijs/broccoli-plugin
[API] Warning: Plugin uses .read/.rebuild API: Filter
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "broccoli-concat": "^3.2.2",
     "broccoli-merge-trees": "^2.0.0",
     "broccoli-stew": "^1.4.2",
-    "broccoli-templater": "^1.0.0",
+    "broccoli-templater": "^2.0.1",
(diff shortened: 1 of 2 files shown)


Turbo87:
The new release requires Node 6+, so unfortunately this has to be considered a breaking change

package.json
```diff
@@ -52,7 +52,7 @@
     "loader.js": "^4.2.3"
   },
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || 8.* || 9.* || >= 10"
```
Turbo87: do we want to support Node 9?

rwjblue: Updated 😸

PR #125, opened 2018-08-16
Rollup dependencies

xg-wang:
rollup dependencies being wrapped for iife format

Previous way of including the built umd from deps directly gives us the unnecessary wrapper which triggers https://github.com/ember-cli/loader.js/issues/60 in some situations

Needs discussion: 
- should we rollup the [babel-plugin-external-helpers](https://github.com/mo/abortcontroller-polyfill/blob/master/rollup.config.js#L32)?
- pre-rollup deps and ship them?

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ const MergeTrees = require('broccoli-merge-trees');
 const concat = require('broccoli-concat');
 const map = stew.map;
 const Rollup = require('broccoli-rollup');
+const babel = require('rollup-plugin-babel');
 
 /*
  * The `index.js` file is the main entry point for all Ember CLI addons.  The
@@ -140,7 +141,13 @@ function treeForBrowserFetch() {
         // abortcontroller is polyfill only, the name is only required by rollup iife
         name: 'AbortController',
         format: 'iife'
-      }
+      },
+      plugins: [
+        babel({
+          babelrc: false,
+          presets: [['env', { modules: false }]]
+        })
+      ]
     }
   });
   // Fork whatwg-fetch to provide umd build before official release, no extra change made.
--- a/index.js
+++ b/index.js
@@ -16,11 +16,14 @@ const path = require('path');
 //   * find - finds files in a tree based on a glob pattern
 //   * map - map content of files in a tree
 //
+// broccoli-rollup: rollup dependencies to expected module format
+//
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -121,26 +124,45 @@ module.exports = {
...
-  const abortcontrollerTree = path.dirname(require.resolve('abortcontroller-polyfill'));
-  const polyfillTree = concat(new MergeTrees([abortcontrollerTree, fetchTree]), {
-    inputFiles: ['abortcontroller-polyfill-only.js', 'fetch.umd.js'],
+  const fetchNode = new Rollup(path.dirname(path.dirname(require.resolve('@xg-wang/whatwg-fetch'))), {
+    rollup: {
+      input: 'fetch.js',
+      output: {
+        file: 'fetch.js',
+        name: 'WHATWGFetch',
+        format: 'iife'
```
asakusuma: Shouldn't we be supplying `amd` instead of `iife` for `treeForVendor()`?

index.js
```diff
@@ -121,26 +124,45 @@ module.exports = {
...
-// polyfill
+// Path to the template that contains the shim wrapper around the browser polyfill
 const templatePath = path.resolve(__dirname + '/assets/browser-fetch.js.t');
 
 
 // Returns a tree containing the browser polyfill (from `whatwg-fetch` and `abortcontroller-polyfill`),
 // wrapped in a shim that stops it from exporting a global and instead turns it into a module
 // that can be used by the Ember app.
 function treeForBrowserFetch() {
+  const abortcontrollerNode = new Rollup(path.dirname(path.dirname(require.resolve('abortcontroller-polyfill'))), {
```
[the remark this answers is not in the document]
xg-wang: Because its `package.json` specifies `"main": "dist/umd-polyfill.js",`, and we need `input: 'src/abortcontroller-polyfill.js'` for rollup option

index.js
xg-wang: Because the exported module for app to do import is `fetch` (ember-fetch) rather than this `fetch` (WHATWGFetch). 
We are exporting `ember-fetch`'s amd module, inside the module loading process of `ember-fetch` we need to load the polyfills, so they should be iife but not a module that exported as amd for app to import.

Issue #98, opened 2018-01-05
Don't work as dependencies


villander:
`ember-fecth` does not work in my consumer app with Fastboot enabled, when I put ember-fetch as `dependencies` in my addon

cosumer app - https://github.com/villander/ember-mvp
addon - https://github.com/villander/ember-apollo-client/blob/apollo-client-2.0/package.json#L34

looks like ember-fetch specifically doesn’t work with fastboot when it’s a nested dependency
because its `updateFastBootManifest()` method is never called

the question is what happens when you wind up calling that hook multiple times?
like if `ember-mvp` and `ember-apollo-client` both have `ember-fetch` installed, does it get called twice?

is it `ember-fetch`’s responsibility to guard against that?

stefanpenner:
@kratiahuja got any cycles?

villander:
@kratiahuja can you help us with that? or some hint of how to solve it, I can work on it if I know the cause

dfreeman:
I opened up https://github.com/ember-fastboot/ember-cli-fastboot/issues/597 to try and start driving toward a solution here. @stefanpenner @kratiahuja would love to get your thoughts

xg-wang:
Close as mentioned in #126 
We can re-open if we can leverage new API for a better solution.

Issue #55, opened 2017-07-12
`Import fetch from 'fetch'` not working w/ 3.2.8


BlueRaja:
Running `ember-fetch` in a Fastboot enabled app, we see the same issue mentioned in https://github.com/ember-cli/ember-fetch/issues/26

Running without Fastboot works fine.

The project is huge so I'm not sure if this is a straight-up an `ember-fetch` issue, or a complex mix of other factors.  We don't have any dependencies relying on `ember-network`.

kratiahuja:
@BlueRaja it would be great if you could provide a minimal reproduction of this issue.

knownasilya:
I see something else, also with 3.2.8

```
ember.debug.js:30911 Error while processing route: dashboard.index FastBoot is not defined ReferenceError: FastBoot is not defined
    at Module.callback (http://stocktonca.mg.dev/engines-dist/dashboard/ember-fetch/fastboot-fetch.js:3:17)
    at Module.exports (http://stocktonca.mg.dev/assets/vendor.js:4574:32)
    at Module._reify (http://stocktonca.mg.dev/assets/vendor.js:4609:59)
    at Module.reify (http://stocktonca.mg.dev/assets/vendor.js:4596:27)
    at Module.exports (http://stocktonca.mg.dev/assets/vendor.js:4572:10)
    at requireModule (http://stocktonca.mg.dev/assets/vendor.js:4498:18)
    at Class._extractDefaultExport (http://stocktonca.mg.dev/assets/vendor.js:214159:20)
    at Class.resolveOther (http://stocktonca.mg.dev/assets/vendor.js:213880:32)
    at Class.superWrapper [as resolveOther] (http://stocktonca.mg.dev/assets/vendor.js:64685:22)
    at Class.resolve (http://stocktonca.mg.dev/assets/vendor.js:29029:35)
```

knownasilya:
Looks like https://github.com/ember-cli/ember-fetch/blob/master/index.js#L84 targets `vendor/*` but the fastboot fetch was moved to `public/`

kratiahuja:
@knownasilya That's the browser version which is wrapped with `if (typeof FastBoot === 'undefined') {..}` check. The Node version of fetch is present in the manifest here: https://github.com/ember-cli/ember-fetch/blob/master/index.js#L91. What's ESA?

From your stack trace it seems sandbox globals in FastBoot aren't being set. Can you put in a minimal reproduction? I would be happy to look into it.

[...]

PR #126, opened 2018-08-17
Add doc for Fastboot nested addon

xg-wang:
For https://github.com/ember-cli/ember-fetch/issues/55 https://github.com/ember-cli/ember-fetch/issues/98.

To encourage host app to install `ember-fetch` seems to be a right thing to do rather than a "workaround".

Any thoughts on this or alternatives? @kratiahuja @stefanpenner

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ export default {
 }
 ```
 
-For addon that supports Fastboot, `ember-fetch` should also be listed as a [peer dependency](https://docs.npmjs.com/files/package.json#peerdependencies). 
-This is because Fastboot only invokes top-level addon's `updateFastBootManifest`, thus `ember-fetch` has to be a top-level addon installed by the host app.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,9 @@ export default {
 }
 ```
 
+For addon that supports Fastboot, `ember-fetch` should also be listed as a [peer dependency](https://docs.npmjs.com/files/package.json#peerdependencies).
```
nlfurniss: Should `For addon that supports ...` be `For apps that support Fastboot...`?

xg-wang: This paragraph is for addon authors, for an addon to support Fastboot it means the app consuming the addon will have no problem to enable Fastboot.

An app can enable Fastboot without `ember-fetch` if none of its addons using `ember-fetch` as well.

PR #128, opened 2018-08-25
Allow relative urls for fastboot and ember-data

rwwagner90:
Copied over the `buildServerUrls` function from ember-fetch-adapter to allow relative urls with fastboot and ember-data.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -1,8 +1,8 @@
 import Ember from 'ember';
 import Mixin from '@ember/object/mixin';
-import { assign, merge } from '@ember/polyfills'
+import { assign, merge } from '@ember/polyfills';
+import { inject as service } from '@ember/service';
 import RSVP from 'rsvp';
-import fetch from 'fetch';
 
 const {
   Logger: { warn }
@@ -87,6 +87,7 @@ export function headersToObject(headers) {
 
   return headersObject;
 }
+
 /**
  * Helper function that translates the options passed to `jQuery.ajax` into a format that `fetch` expects.
  * @param {Object} _options
@@ -97,7 +98,7 @@ export function mungOptionsForFetch(_options, adapter) {
   // This allows this mixin to be backward compatible with Ember < 2.5.
   const combineObjs = (assign || merge);
   const options = combineObjs({
-    credentials: 'same-origin',
+    credentials: 'same-origin'
   }, _options);
 
   let adapterHeaders = adapter.get('headers');
@@ -134,6 +135,7 @@ export function mungOptionsForFetch(_options, adapter) {
 
   return options;
 }
+
 /**
  * Function that always attempts to parse the response as json, and if an error is thrown,
  * returns an object with 'data' set to null if the response is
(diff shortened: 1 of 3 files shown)


addon/mixins/adapter-fetch.js
```diff
@@ -165,7 +169,20 @@ export function determineBodyPromise(response, requestData) {
 }
 
 export default Mixin.create({
-  adapter: service(),
+  fastboot: computed(function() {
+    let owner = getOwner(this);
+    return owner && owner.lookup('service:fastboot');
+  }),
+
+  protocol: computed(function() {
```
nlfurniss: why use CPs with dependent keys? Can these be set in `init`?

rwwagner90:
@nlfurniss I copied this stuff from https://github.com/tchak/ember-fetch-adapter/blob/master/addon/index.js

It allows users of this mixin to make requests when in fastboot. Without these changes, you get an error with node-fetch complaining about `only absolute URLs are supported`.

rwwagner90:
I have been speaking with @tchak about this. Might be good for him to weigh in here 😃

rwwagner90:
@nlfurniss @tchak any updates on this?

chrism:
I've been getting this error too when using with Fastboot after following the instructions.

`Error while processing route: index Only absolute URLs are supported TypeError: Only absolute URLs are supported`

It would be great to see this merged.

rwwagner90:
@nlfurniss @rwjblue what do we need to do to get this merged?

nlfurniss:
Can you rebase and resolve the conflicts? Also what about tests?

rwwagner90:
@nlfurniss yes, I can rebase, but tests would need a very special setup to force fastboot and ember-data, and I am not sure it would be something we could easily do. If I rebase, can we merge as is? This has been open quite awhile.

addon/mixins/adapter-fetch.js
```diff
@@ -251,13 +271,38 @@ export default Mixin.create({
...
- * Allows for the error to be selected from either the
- * response object, or the response data.
- * @param {Object} response
- * @param {Object} payload
- */
+    if (protocolRelativeRegex.test(url)) {
+      return `${protocol}${url}`;
+    } else if (!httpRegex.test(url)) {
+      try {
+        return `${protocol}//${host}${url}`;
```
xg-wang: When will this throw?

rwwagner90:
> I copied this stuff from https://github.com/tchak/ember-fetch-adapter/blob/master/addon/index.js
> 
> It allows users of this mixin to make requests when in fastboot. Without these changes, you get an error with node-fetch complaining about `only absolute URLs are supported`.

@xg-wang I just copied all this, as mentioned previously. Since I know it works currently, and it's copied from this other addon, I am not sure what the best way to proceed is.

addon/mixins/adapter-fetch.js
xg-wang: @chrism Yea you're absolutely right, thansk for pointing out!

xg-wang:
Close by https://github.com/ember-cli/ember-fetch/pull/143

PR #134, opened 2018-10-21
Deprecate Logger and merge

xg-wang:
Deprecations
- Ember.Logger(RFC297)
- Ember.merge(RFC340)

ember-try version compatibility now covers back to ember2.3
@rwjblue The [deprecation guide](https://emberjs.com/deprecations/v3.x#toc_ember-polyfills-deprecate-merge) says `Ember.assign` is not available in Ember <= 2.4, but I found it in [old release branches](https://github.com/emberjs/ember.js/blob/lts-2-4/packages/ember-metal/lib/assign.js). Am I missing anything here?

Also use whatwg-fetch instead of forked version. Breaking changes include:
- [unsupported body type](https://github.com/github/fetch/commit/08bc7423c9d26dd9f21905fee3b83660723e2d37)
- [Change `credentials` default value to `same-origin`](https://github.com/github/fetch/commit/048626722124b45ed7bc58ec841822610f08798b)

Related PR https://github.com/pretenderjs/pretender/pull/235

This can be a major bump with https://github.com/ember-cli/ember-fetch/pull/130#discussion_r216174706

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,9 @@ module.exports = {
   env: {
     browser: true
   },
-  rules: {},
+  rules: {
+    'no-console': ["error", { allow: ['warn'] }]
+  },
   overrides: [
     // node files
     {
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ env:
   matrix:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
-    - EMBER_TRY_SCENARIO=ember-lts-2.12
     - EMBER_TRY_SCENARIO=ember-lts-2.16
     - EMBER_TRY_SCENARIO=ember-lts-2.18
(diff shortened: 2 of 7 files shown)


nlfurniss:
`The deprecation guide says Ember.assign is not available`

where does it say that?

xg-wang:
@nlfurniss Sorry, I meant `Ember <= 2.4`

nlfurniss:
Why do we want to cover back to 2.3?

xg-wang:
The compatible versions were not specified explicitly. But I think it would be nice to have for user to know about.

It was when removing
```
// This allows this mixin to be backward compatible with Ember < 2.5.
const combineObjs = (assign || merge);
```
I wanted to verify the oldest ember version we are actually supporting.

rwjblue:
ember-assign-polyfill specifically exists for this reason

rwjblue:
However, we should strongly consider officially dropping support for Ember < 2.16...

Issue #72, opened 2017-09-11
Typescript definitions don't work


dwickern:
If you `import fetch from 'fetch'`, typescript won't resolve it. 
If you `import fetch from 'ember-fetch'`, typescript _will_ resolve it, but you'll have a runtime error.

When you import `fetch`, typescript looks in `node_modules/fetch`. It won't scan the whole `node_modules` directory for anyone exporting a `fetch` module. You can work around this with an explicit path mapping in `tsconfig.json`, but it's not conventional.

Maybe this package could export `ember-fetch`, and eventually deprecate `fetch`?

/cc @toranb @chriskrycho

chriskrycho:
Ugh. I was hoping that `fetch` would be an actual npm package, but it's `whatwg-fetch`, which isn't even a little bit better than `ember-fetch`. I agree that the best option here is exporting from `ember-fetch` instead, since the *real* npm fetch package [is something *totally* different](https://www.npmjs.com/package/fetch).

Your alternative does bring up an interesting point I'd like to think about in *other* packages as makes sense: if there's a *good reason* to, we can presumably teach people how to update the `tsconfig.json` as part of the addon installation process. That should *certainly* be a last-resort option, though.

toranb:
I was actually curious about the (most) correct path forward myself and thought it might be worth asking @DanielRosenwasser to chime in :)

warning and type defs are below for quick reference

`A default export can only be used in an ECMAScript-style module.`

```ts
export = fetch;

declare namespace fetch {
  export default function fetch(input: RequestInfo, init?: RequestInit): Promise<Response>;
}
```

DanielRosenwasser:
> A default export can only be used in an ECMAScript-style module.

Right now by default, TypeScript assumes the mapping between ES modules and CJS modules to be 1:1 (the namespace record and the imported object are treated the same). To get around that, you can use `--allowSyntheticDefaultImports`.

>  You can work around this with an explicit path mapping in tsconfig.json, but it's not conventional.

An alternative idea is to add

```json
    "fetch": "ember-cli/ember-fetch"
```

which will get the current state of `ember-fetch` from this GitHub repo as a dependency

toranb:
@DanielRosenwasser thanks for the detailed reply! Reading this it would appear the synthetic option is most appropriate. Any downside to using this option generally speaking?

toranb:
[...]

PR #136, opened 2018-10-28
TypeScript support

xg-wang:
~Document TypeScript usage as suggested by @DanielRosenwasser at https://github.com/ember-cli/ember-fetch/issues/72#issuecomment-365407346.~

~Provide a build time detection for `tsconfig.json` as a hint user needs the type support.~

This PR installs `ember-cli-typescript` and add typescript support doc to README.

Thanks @mike-north for the discussing the type resolution and detection strategy.


index.d.ts
```diff
@@ -1,9 +1,27 @@
-// Type definitions for Ember Fetch
+// Type definitions for ember-fetch
 // Project: https://github.com/ember-cli/ember-fetch
-// Definitions by: Toran Billups <https://github.com/toranb>
-// TypeScript Version: 2.3
+// Definitions by: Toran Billups <https://github.com/toranb>, Thomas Wang<https://github.com/xg-wang>
+/// <reference lib="dom"/>
 
 declare module 'fetch' {
```
mike-north:
```suggestion
```

index.d.ts
```diff
@@ -1,9 +1,27 @@
...
+    prototype: Response;
+    new(body?: BodyInit | null, init?: ResponseInit): Response;
+    error(): Response;
+    redirect(url: string, status?: number): Response;
+  };
+  export const AbortController: {
+    prototype: AbortController;
+    new(): AbortController;
+  };
+}
```
mike-north:
```suggestion

```

index.d.ts
```diff
@@ -1,9 +1,27 @@
-// Type definitions for Ember Fetch
+// Type definitions for ember-fetch
 // Project: https://github.com/ember-cli/ember-fetch
-// Definitions by: Toran Billups <https://github.com/toranb>
-// TypeScript Version: 2.3
+// Definitions by: Toran Billups <https://github.com/toranb>, Thomas Wang<https://github.com/xg-wang>
+/// <reference lib="dom"/>
```
mike-north:
```suggestion
/// <reference lib="es2015" />
/// <reference lib="dom" />
```

mike-north:
@xg-wang I filed a bug for the NPM issue we ran into https://npm.community/t/npm-install-for-package-with-local-dependency-fails/754/14

xg-wang:
Great to know there's a pending PR in npm to fix the issue (which I'll document more about when this PR proceeds).
https://github.com/npm/cli/pull/3

addon/mixins/adapter-fetch.ts
```diff
@@ -0,0 +1,187 @@
...
+  let headersObject: PlainObject = {};
+
+  if (headers) {
+    headers.forEach((value, key) => (headersObject[key] = value));
+  }
+
+  return headersObject;
+}
+
+export default Mixin.create({
```
xg-wang: I wanted to do `Mixin.create<TheMixinTypeLiteral, DS.RestAdapter>` here. But it seems it's not correct. For example `this.ajaxOptions` might be wrapped by base adapter's `this.ajaxOptions`'s `this._super(...arguments)` and return not necessarily `FetchOptions`.

xg-wang:
@mike-north can you review again?

README.md
```diff
@@ -35,6 +35,12 @@ Available imports:
 import fetch, { Headers, Request, Response, AbortController } from 'fetch';
 ```
 
+### Use with TypeScript
+To use `ember-fetch` with TypeScript or enable editor's type support, add `"fetch": "ember-cli/ember-fetch"` to your app's `devDependencies`.
+This will get the current state of `ember-fetch` from this GitHub repo as a dependency.
```
mike-north: I'm not sure we want to advise that people do this, since they'll end up tracking the master branch rather than SemVer versioned releases.

README.md
```diff
@@ -35,6 +35,12 @@ Available imports:
...
+
+You can also add `"fetch": ["node_modules/ember-fetch"]` to your `tsconfig.json`.
```
mike-north: Please specify _where_ this should be added in `tsconfig.json`

addon/ajax.ts
```diff
@@ -0,0 +1,14 @@
+
+import fetch from 'fetch';
+
+export default async function ajax(
+  input: RequestInfo,
+  init?: RequestInit
+): Promise<Response> {
+  return fetch(input, init).then(response => {
```
mike-north: I would either

A) leave `ajax` as an `async` function, and use `await` to wait for promises instead of `.then`
B) treat `ajax` as a regular (non-`async`) function and use `.then` to wait for promises instead of `await`

Currently you're mixing (A) and (B) by making the function `async` but never `await`ing

addon/mixins/adapter-fetch.ts
mike-north: from a typescript standpoint, there's no "clean" way to declare assumptions about the type of object the mixin will be applied to (this is one of the general pitfalls of using mixins in general).

The best you could do is explicitly specify `this` in your method signatures. For example

```ts
ajaxOptions(this: DataAdapter, url: string, type: Method, options: object): FetchOptions {
  // in here `this` will be a DataAdapter
  ...
} 

```

addon/mixins/adapter-fetch.ts
```diff
@@ -0,0 +1,187 @@
...
+   */
+  _fetchRequest(url: string, options: RequestInit): RSVP.Promise<Response> {
+    return fetch(url, options);
+  },
+
+  /**
+   * @override
+   */
+  ajaxSuccess(
+    adapter: any,
```
mike-north: Can we do better than `any` here? maybe a `DS.Adapter`?

addon/mixins/adapter-fetch.ts
```diff
@@ -0,0 +1,187 @@
...
+    response: Response,
+    payload: Nullable<string | object>,
+    requestData: { url: string; method: string }
+  ): object | DS.AdapterError | RSVP.Promise<never> {
```
mike-north: Can we see of `RSVP.Promise<void>` works here?

addon/types.ts
```diff
@@ -0,0 +1,36 @@
+import Mix from '@ember/polyfills/types';
+
+export type Nullable<T> = T | null | undefined;
+
+export interface PlainObject {
+  [key: string]: string | PlainObject | PlainObject[];
+}
+
+export interface PlainHeaders {
+  [key: string]: string | undefined | null;
```
mike-north:
```suggestion
  [key: string]?: string | null;
```

addon/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  headers?: PlainHeaders;
+};
+
+export type FetchOptions = Mix<
+  AjaxOptions,
+  { body?: BodyInit | null; method?: Method }
+>;
+
+export function isPlainObject(obj: any): obj is PlainObject {
+  return Object.prototype.toString.call(obj) === '[object Object]';
```
mike-north: this isn't really the right implementation for this guard, since it will pass for lots of objects that aren't `PlainObject`. For example `{ foo() { return 'bar'; } }` will still coerce to string as `[object Object]`

addon/utils/mung-options-for-fetch.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export default function mungOptionsForFetch(
+  options: AjaxOptions
+): FetchOptions {
+  const hash = assign(
+    {
+      credentials: 'same-origin'
+    },
+    options
+  ) as FetchOptions;
```
mike-north: Do you remember why we end up doing this, instead of adding `credentials` to the `FetchOptions` type?

addon/utils/serialize-query-params.ts
```diff
@@ -1,15 +1,17 @@
...
  * This is directly copied from jQuery.param.
- * @param {Object} queryParamsObject
- * @returns {String}
  */
-export function serializeQueryParams(queryParamsObject) {
-  var s = [];
+export function serializeQueryParams(
+  queryParamsObject: object | string
+): string {
+  var s: any[] = [];
```
mike-north:
```suggestion
  const s: any[] = [];
```

types/dummy/index.d.ts
```diff
@@ -0,0 +1 @@
+
```
mike-north: revert unnecessary whitespace

README.md
xg-wang: I tend to document `tsconfig.json` only.

addon/ajax.ts
xg-wang: oooh I was in the middle of transforming it to async but decide to leave it as is, because we are going to have a new [service](https://github.com/emberjs/ember.js/issues/17476) and probably deprecate this.

addon/mixins/adapter-fetch.ts
xg-wang: > The best you could do is explicitly specify `this` in your method signatures.

We don't need this, `this` has a type here already
```js
static create<T, Base = EmberObject>(
    args?: MixinOrLiteral<T, Base> & ThisType<Fix<T & Base>>
): Mixin<T, Base>;
```

This is `Mixin.create` signature, so technically I can write `Mixin.create(FetchAdapter, DS.RESTAdapter>`

The issue is that

1. I have to manually define the object literal type
2. There might be **other** mixins, so type of `this` will still be inaccurate at runtime.

But I think it's still better than doing `adapter: any` here. Updated it to have an interface.

addon/mixins/adapter-fetch.ts
xg-wang: `reject(reason?: any, label?: string | undefined): RSVP.Promise<never>`
It's typed for `RSVP.rejct`

addon/types.ts
xg-wang: I think typescript doesn't let me do this.

addon/utils/mung-options-for-fetch.ts
xg-wang: So I need to return a `FetchOptions` type required by ts, `assign` doesn't automatically gives me this kind of type. Explicit declaration is needed. `credentials` is an optional field in `fetch` API, but we 100% should add this to `FetchOptions` type. Updated.

addon/utils/serialize-query-params.ts
xg-wang: There're lots of `var` here because it was copy-pasted from jquery code. We can do a refactor in the future and leave this PR a type only one.

addon/types.ts
xg-wang: I want to preserve all current behaviors so it be a type only PR. (I assume `Object.prototype.toString.call(obj)` equals `String(obj)`. There exists ways to break this but we can leave it in another fix PR.

addon/types.ts
mike-north: you'll have to change it to a `type` instead of `interface`

```ts
interface Foo{
    bar: string;
    baz: number;
}

type A<T> = { [K in keyof T]?: T[K] | null }

type X = A<Foo>;/*
{
    bar?: string | null | undefined;
    baz?: number | null | undefined;
}
*/

```
https://typescript-play.js.org/#code/JYOwLgpgTgZghgYwgAgGIHt0G8BQz-IBGcUAXMgM5hSgDmA3HgcQF7kgCuAtodIwL44cYAJ4AHFAEEAPABUAfMgC8yLMgDaAaWShkAawgj0MZLIC6AfnKytZ5AB9knADbPkg4eJQANZchkY6PL0APQAVDi4BEQkVpTUdA5OHK5JHCAAJhAwoBAZjNGscZw80Ekubo7pWTkgeQI4YSE4QA

README.md
mike-north: please clarify. This pertains to `package.json`, not `tsconfig.json`

> add `"fetch": "ember-cli/ember-fetch"` to your app's `devDependencies`.
This will get the current state of `ember-fetch` from this GitHub repo as a dependency.

addon/mixins/adapter-fetch.ts
mike-north: `Promise<never>` is assignable to `Promise<void>` so you should be ok to make this change

```ts
function handlePromise(p: Promise<void>): void {

}
handlePromise(new Promise<never>(null as never))
```
https://typescript-play.js.org/#code/GYVwdgxgLglg9mABACwIZgCYBsCmAFAJzgFsYBnHACgAcAuRQk8nAHgDc4YMA+ASno5dEAbwBQogL6i0mXI1IVKYHAHcGRBa2VscBbkpBYsiVGUTbdvXkA

addon/types.ts
xg-wang: I changed `PlainHeaders` and `PlainObject` from `interface` to `type`, and made `PlainObject` more generic.

README.md
xg-wang: I removed the `"fetch": "ember-cli/ember-fetch"` instruction and only left a `tsconfig` example

mike-north:
Looks like you've responded to all my feedback 👍. This looks good to me!

PR #139, opened 2018-10-30
Set body to undefined for 204/205/HEAD responses

nlfurniss:
This brings the mixin inline with jQuery.ajax, which correctly returns `undefined` for the body of a 204/205/HEAD response/request since those are not allowed to contain a body.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -65,9 +65,8 @@ export function mungOptionsForFetch(_options, adapter) {
 }
 /**
  * Function that always attempts to parse the response as json, and if an error is thrown,
- * returns an object with 'data' set to null if the response is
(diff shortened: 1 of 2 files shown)


addon/mixins/adapter-fetch.js
```diff
@@ -82,7 +82,7 @@ export function determineBodyPromise(response, requestData) {
       }
       const status = response.status;
       if (response.ok && (status === 204 || status === 205 || requestData.method === 'HEAD')) {
```
xg-wang: https://github.com/jquery/jquery/blob/2.2-stable/src/ajax.js#L751-L764
jQuery is reading `response.data` if not 204 | 304 | HEAD. Can you verify?

nlfurniss: Sorry, I don't get what you're asking

nlfurniss: This doesn't seem to be an issue. We can revisit if it becomes ones

PR #140, opened 2018-11-01
Fix typo

xg-wang:
statusTest -> statusText


xg-wang:
I'll look into the test cleanup later today when I have time for it.

PR #164, opened 2018-11-16
Improve ESLint setup

Turbo87:
see commit messages

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "repository": "https://github.com/ember-cli/ember-fetch",
   "scripts": {
     "build": "ember build",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   "repository": "https://github.com/ember-cli/ember-fetch",
   "scripts": {
     "build": "ember build",
-    "lint:js": "eslint ./*.js addon addon-test-support app config lib server test-support tests",
+    "lint:js": "eslint . --cache",
```
buschtoens: Nice! But we should `.eslintcache` to `.gitignore` then, I guess?

Turbo87: you're totally right, I forgot that 🤦‍♂️ 

fixed!

Issue #166, opened 2018-11-17
Could not find module `fetch/setup` imported from `prog/instance-initializers/setup-fetch`


lougreenwood:
Hello, I'm getting the following build error on my CI server. Downgrading to 5.1 (before `/instance-initializers/setup-fetch` was introduced) fixes the issue.

```
Build Error (prember)

Could not find module `fetch/setup` imported from `prog/instance-initializers/setup-fetch`

=================================================================================

ENV Summary:

  TIME: Sat Nov 17 2018 19:26:34 GMT+0000 (UTC)
  TITLE: ember
  ARGV:
  - /usr/bin/node
  - /usr/local/bin/ember
  - b
  - -prod
  EXEC_PATH: /usr/bin/node
  TMPDIR: /tmp
  SHELL: /bin/ash
  PATH:
  - /usr/local/sbin
  - /usr/local/bin
  - /usr/sbin
  - /usr/bin
  - /sbin
  - /bin
  PLATFORM: linux x64
  FREEMEM: 9050533888
  TOTALMEM: 73765941248
  UPTIME: 78321
  LOADAVG: 8.33837890625,5.37841796875,3.8583984375
  CPUS:
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
[...]

PR #167, opened 2018-11-19
Fix fetch public/fastboot-fetch.js module definition for Fastboot

xg-wang:
Fix https://github.com/ember-cli/ember-fetch/issues/166, depends on https://github.com/ember-cli/ember-fetch/pull/171

In v6.2.0 `fetch/setup` was added to `public/fastboot-fetch.js` so we can setup `fetch` module with fastboot request info. However this file will be overwritten by any addon's dependency `ember-fetch` based on last-write-win. To fix this:

- Rename public/fastboot-fetch.js to public/fetch-fastboot.js to avoid
lower version ember-fetch overwrite this file.
- Overrides treeForPublic in index.js to only include public tree if top
level addon.

For example, top level addon `ember-fetch` is 6.2.1, transitive dependency is anything before 6.2.0 will result in this dist:

```
dist
├── assets
│   ├── ...
├── ember-fetch
│   ├── fastboot-fetch.js
│   └── fetch-fastboot.js
```

6.2.1 as top level addon will `manifest.vendorFiles.push('ember-fetch/fetch-fastboot.js');` so the `fastboot.js` will just be ignored.

This commit also defines fetch module with a `setupFastboot` method to set host and
protocol for every instance. This could avoid module redefinition.


Turbo87:
What if ember-fetch is never the top level addon? What if it is only used as a subdep with several versions?

xg-wang:
@Turbo87 `updateFastBootManifest` is only called on top level addons. For subdeps only `treeForVendor` has effect, and we don't calculate cache key for `vendor`

Turbo87:
@xg-wang this PR seems to contain a lot of unrelated changes now. can you make sure to open dedicated PRs for those unrelated changes?

xg-wang:
@Turbo87 This PR is now only relying on https://github.com/ember-cli/ember-fetch/pull/171.

samselikoff:
#171 is merged, this good to go?

runspired:
We may want to rethink this design a bit now that `ember-fetch` is included as a dependency by `ember-data`

rwjblue:
Agree

dnalagatla:
+1

e-monson:
> We may want to rethink this design a bit now that `ember-fetch` is included as a dependency by `ember-data`

What are the implications of `ember-fetch` being included as a dependency by `ember-data`? Just wondering if it could be related to my issue (#442 )

PR #172, opened 2018-11-27
Make configuration work in engines

mydea:
This makes sure that all nested dependencies use the correct application configuration, and that the import/configuration works in nested addons, especially in engines.

This fixes #133.

I tried it in an app of ours with engines & nested dependencies to ember-fetch, and this ensures that it always got the correct options passed.

Note: While investigating this, I also found out that it wasn't even taking the correct options for building. E.g. If, in my engines dummy app, I had `preferNative: true` specified in the `ember-cli-build.js`, it never actually got to this value, instead reading the config from ember-fetch's config (somehow? not entirely sure).

PR #173, opened 2018-11-27
Do not include polyfill if browser targets don't need it

mydea:
This is my stab at fixing #45.

If the target browsers all support native fetch, this will avoid importing the polyfill(s) and just setup the exports, so that you can continue using `import fetch from 'fetch';` with no code changes.

By default, it will use your specified target browsers, but you can also override this in the options, e.g.:

```js
'ember-fetch': {
  preferNative: true,
  browers: ['ie 11']
}
```

If no browsers are found at all, it will always include the polyfill.

This PR is on top of #172 , as I could not get to the configuration via `this.buildConfig` in the `treeForBrowserFetch` method.

If you have any suggestions for improvements, I'd be happy to try to implement them - To be honest, it was a bit tricky to implement this, so there might be better ways to do some things.

As a sidenote, I wasted some time until I figured out that the polyfill is included in development environment even if it shouldn't (due to the browser targets), as `ember-cli-pretender` will _also_ include the polyfill in development/test environment. So as long as you have `ember-cli-pretender` installed, you'll need to actually run `ember s --prod` to get a build without the polyfill (if you remove IE 11 from the targets for production).

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ otherwise the polyfilled `fetch` will be installed during the first pass of the 
 
 If set to `false`, the polyfilled `fetch` will replace native `fetch` be there or not.
 
-If all your browser targets (defined in `config/targets.js`) support native `fetch`, and `preferNative: true`, the polyfill will not be included in the output build. If, for some reason, you still need the polyfill to be included in the bundle, you can set `alwaysIncludePolyfill: true`.
+If all your [browser targets](https://guides.emberjs.com/release/configuring-ember/build-targets/) support native `fetch`, and `preferNative: true`, the polyfill will not be included in the output build. If, for some reason, you still need the polyfill to be included in the bundle, you can set `alwaysIncludePolyfill: true`.
 
 The way you do import remains same.
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ otherwise the polyfilled `fetch` will be installed during the first pass of the 
 
 If set to `false`, the polyfilled `fetch` will replace native `fetch` be there or not.
 
+If all your browser targets (defined in `config/targets.js`) support native `fetch`, and `preferNative: true`, the polyfill will not be included in the output build. If, for some reason, you still need the polyfill to be included in the bundle, you can set `alwaysIncludePolyfill: true`.
+
 The way you do import remains same.
 
 ## Browser Support
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -33,6 +33,10 @@
 
     <%= moduleBody %>
 
+    if (!self.fetch) {
+      throw new Error('fetch is not defined - maybe your browser targets are not covering everything you need?');
+    }
+
     var pending = 0;
     function decrement(result) {
       pending--;
--- a/index.js
+++ b/index.js
@@ -74,8 +74,13 @@ module.exports = {
       importTarget = this;
     }
 
-    const options = app.options['ember-fetch'] || { preferNative: false };
-    options.browsers = options.browsers || this.project.targets && this.project.targets.browsers;
+    const options = app.options['ember-fetch'] || {};
+    const { preferNative = false, alwaysIncludePolyfill = true } = options;
+
+    options.browsers = this.project.targets && this.project.targets.browsers;
+    options.preferNative = preferNative;
+    options.alwaysIncludePolyfill = alwaysIncludePolyfill;
+
     app._fetchBuildConfig = options;
 
     importTarget.import('vendor/ember-fetch.js', {
@@ -117,7 +122,8 @@ module.exports = {
 
     const app = this._findApp();
     // If the polyfill is not included, we force the preferNative behavior
-    const preferNative = app._fetchBuildConfig.preferNative || this._checkSupports('fetch');
+    const options = app._fetchBuildConfig;
+    const preferNative = options.preferNative || (!options.alwaysIncludePolyfill && this._checkSupports('fetch'));
 
     return debug(map(browserTree, (content) => `if (typeof FastBoot === 'undefined') {
       var preferNative = ${preferNative};
@@ -148,7 +154,10 @@ module.exports = {
   // wrapped in a shim that stops it from exporting a global and instead turns it into a module
   // that can be used by the Ember app.
   treeForBrowserFetch() {
-    const needsFetchPolyfill = !this._checkSupports('fetch');
+    const app = this._findApp();
+    const alwaysIncludePolyfill = app._fetchBuildConfig.alwaysIncludePolyfill;
+    const needsFetchPolyfill = alwaysIncludePolyfill || !this._checkSupports('fetch');
+    const needsAbortControllerPolyfill = alwaysIncludePolyfill || !this._checkSupports('abortcontroller');
 
     const abortcontrollerNode = debug(new Rollup(path.dirname(path.dirname(require.resolve('abortcontroller-polyfill'))), {
       rollup: {
@@ -173,15 +182,23 @@ module.exports = {
       }
     }), 'whatwg-fetch');
 
-    const polyfillNode = debug(concat(new MergeTrees([abortcontrollerNode, fetchNode]), {
-      inputFiles: ['abortcontroller.js', 'fetch.js'],
+    let inputNodes = [abortcontrollerNode, fetchNode];
+    let inputFiles = ['abortcontroller.js', 'fetch.js'];
+
+    if (!needsFetchPolyfill && needsAbortControllerPolyfill) {
+      inputNodes = [abortcontrollerNode];
+      inputFiles = ['abortcontroller.js'];
+    }
+
+    const polyfillNode = debug(concat(new MergeTrees(inputNodes), {
+      inputFiles,
       outputFile: 'ember-fetch.js',
       sourceMapConfig: { enabled: false }
     }), 'after-concat');
 
     return debug(new Template(polyfillNode, TEMPLATE_PATH, function(content) {
       return {
-        moduleBody: needsFetchPolyfill ? content : ''
+        moduleBody: needsFetchPolyfill || needsAbortControllerPolyfill ? content : ''
       };
     }), 'browser-fetch');
   },
--- a/index.js
+++ b/index.js
@@ -74,12 +74,8 @@ module.exports = {
       importTarget = this;
     }
 
-    const options = app.options['ember-fetch'] || {};
-    const { preferNative = false, alwaysIncludePolyfill = true } = options;
-
+    const options = Object.assign({ preferNative: false, alwaysIncludePolyfill: true}, app.options['ember-fetch'] || {});
     options.browsers = this.project.targets && this.project.targets.browsers;
-    options.preferNative = preferNative;
-    options.alwaysIncludePolyfill = alwaysIncludePolyfill;
 
     app._fetchBuildConfig = options;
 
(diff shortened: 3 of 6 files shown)


xg-wang:
Why do we need the `browsers` option if we already have target.js?

mydea:
I basically lifted this from ember-cli-autoprefixer - it also makes testing a bit easier. In theory, it could also be used for special needed behavior, e.g. forcing including of the polyfill even though your targets don't need it. But I can remove the option if so desired :)

mydea:
So, do you want me to remove the `browsers` option, or should I leave it in?

assets/browser-fetch.js.t
```diff
@@ -33,6 +33,10 @@
 
     <%= moduleBody %>
 
+    if (!self.fetch) {
```
xg-wang: This error should be changed to provide more actionable info if we agree on changes below.

index.js
```diff
@@ -113,7 +116,8 @@ module.exports = {
     }
 
     const app = this._findApp();
-    const preferNative = app._fetchBuildConfig.preferNative;
+    // If the polyfill is not included, we force the preferNative behavior
+    const preferNative = app._fetchBuildConfig.preferNative || this._checkSupports('fetch');
```
xg-wang: I think we should respect `preferNative` over browser support and `alwaysIncludePolyfill`. The reason is people may be relying some xhr features and don't want native fetch even it exists.

index.js
```diff
@@ -131,6 +135,8 @@ module.exports = {
   // wrapped in a shim that stops it from exporting a global and instead turns it into a module
   // that can be used by the Ember app.
   treeForBrowserFetch() {
+    const needsFetchPolyfill = !this._checkSupports('fetch');
```
xg-wang: The issue here is we don't just polyfill `fetch`, we polyfill `AbortController` as well. And they have different browser support. `caniuse also has a [note](https://caniuse.com/#search=fetch)

> Safari has window.AbortController defined in the DOM but it's just a stub, it does not abort requests at all. The same issue also affects Chrome on IOS and Firefox on IOS because they use the same WebKit rendering engine as Safari.

which is detected by our polyfill lib [programmatically](https://github.com/mo/abortcontroller-polyfill/commit/256fde6b551678bbb72a7821c1fdbb734bfbe84a). I'm only brain debugging this... so apparent we need to verify API response or we come up with a better way.

But I totally agree if `preferNative && supportNativeFetchAndAbortController && !alwaysIncludePolyfill` we should just strip off the polyfill.

Or if apart from `preferNative` and `alwaysIncludePolyfill` we can provide 2 separate group of these for `fetch` and `abortController` with reasonable defaults. User can have finer-grained control of this addon. What do you think?

index.js
```diff
@@ -73,7 +74,9 @@ module.exports = {
       importTarget = this;
     }
 
-    app._fetchBuildConfig = app.options['ember-fetch'] || { preferNative: false };
+    const options = app.options['ember-fetch'] || { preferNative: false };
+    options.browsers = options.browsers || this.project.targets && this.project.targets.browsers;
```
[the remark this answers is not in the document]
mydea: Sure! So I'll add an `alwaysIncludePolyfill` option - I guess that should default to `true` for now, to make this non-breaking under all circumstances?

index.js
mydea: But if we do not include the polyfill, we need to set preferNative to true - otherwise, it will fail.
If we have two options: `preferNative` (default false) and `alwaysIncludePolyfill` (default true). 

Possible combinations:

`{ preferNative: true, alwaysIncludePolyfill: true }` --> polyfill is always included, but might not be used.

`{ alwaysIncludePolyfill: false }` --> In this case, the polyfill might be loaded or not loaded (it might either load fetch & abortcontroller polyfills, or only the abortcontroller polyfill, or none). If at least one polyfill is NOT loaded, then `preferNative` has to be `true`, otherwise the unpolyfilled parts will not work.

index.js
mydea: True, I did not think about this - Can I Use has just recently updated this to correctly state that abortcontroller does not work in Safari. When I looked the last time, it seemed as if all fetch-supporting browsers also supported abortcontroller - which is clearly not the case.

We can just load the necessary polyfills - e.g. fetch & abortcontroller, or only abortcontroller, or none!

mydea:
So I've pushed another commit, checking `AbortController` support separately from `fetch` support. 
This includes a new `alwaysIncludePolyfill` option, which defaults to `true` (e.g. the previous behavior - which might be changed in an upcoming major version I guess?). If this is `true`, it will always include the `fetch` and the `AbortController` polyfills.

If this is set to false, it will check if all browser targets support `fetch` and/or `AbortController`, and include only the necessary polyfills based on browser support. If at least one of the polyfills is _not_ included, it will automatically set `preferNative` to true to ensure the non-polyfilled code works as expected.

index.js
xg-wang: I was saying the logic should be inverse: if the `preferNative` is set to `true`, then check second option `alwaysIncludePolyfill` and browser supports.

The behavior would be 
- `{ preferNative: true, alwaysIncludePolyfill: true }`  --> polyfills are always included, but might not be used
- `{ preferNative: true, alwaysIncludePolyfill: false }` --> conditionally include polyfills based on browser support
- `{ preferNative: false }` --> __skip__ checking `alwaysIncludePolyfill` and caniuse API request, always include polyfills

index.js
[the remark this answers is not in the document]
xg-wang: Agree it should default to `true` which preserves all current behaviors. 
Setting it to `false` isn't breaking in the final output perspective. But making API requests during build time is.

index.js
```diff
@@ -167,19 +182,39 @@ module.exports = {
...
+    const polyfillNode = debug(concat(new MergeTrees(inputNodes), {
+      inputFiles,
       outputFile: 'ember-fetch.js',
       sourceMapConfig: { enabled: false }
     }), 'after-concat');
 
     return debug(new Template(polyfillNode, TEMPLATE_PATH, function(content) {
       return {
-        moduleBody: content
+        moduleBody: needsFetchPolyfill || needsAbortControllerPolyfill ? content : ''
```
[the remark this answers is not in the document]
mydea: Not sure if I understand correctly, but it doesn't work with passing an empty array of nodes, you need to pass at least one node.

index.js
mydea: Ah, got it.
I guess the question is, does it make sense for `alwaysIncludePolyfill` to default to true then, really? I mean, I guess it is more of an edge case to say: "I want to use the native polyfill when possible, but still always include the polyfill file"? Because now, to get this (IMHO most sensible) outcome, you need to set:

```
'ember-fetch': {
  preferNative: true,
  alwaysIncludePolyfill: false
}
```

Which feels a bit counter-intuitive - as these two options are pretty entangled and depend on each other in difficult to understand ways (e.g. setting `alwaysIncludePolyfill` does nothing if `preferNative=false`, ...). So I guess, wouldn't it be easier to understand to make `alwaysIncludePolyfill` default to false, so you can still, if needed for whatever edge-case reason, opt out of it, but the "happy path" is much simpler by just setting `preferNative: true`?

test/browsers-target-test.js
```diff
@@ -0,0 +1,182 @@
...
+const co = require('co');
+
+describe(`Do not include the polyfill if the browser targets match`, function() {
+  let output, subject, addon;
+
+  beforeEach(function() {
+    addon = Object.create(AddonFactory);
+    Object.assign(addon, {
+      addons: [],
+      _fetchBuildConfig: {
```
xg-wang: better also include `alwaysIncludePolyfill` as it has a default value

test/browsers-target-test.js
```diff
@@ -0,0 +1,182 @@
...
+    });
+    subject = addon.treeForVendor();
+    output = helpers.createBuilder(subject);
+  });
+
+  afterEach(co.wrap(function* () {
+    yield output.dispose();
+  }));
+
+  it('preferNative is built into vendor file', co.wrap(function* () {
```
xg-wang: The description message doesn't make sense.

test/browsers-target-test.js
```diff
@@ -0,0 +1,182 @@
...
+    yield output.build();
+    let files = output.read();
+    expect(files).to.have.all.keys('ember-fetch.js');
+    expect(files['ember-fetch.js']).to.include(`var preferNative = true`);
```
xg-wang: The result is not correct.

This is my debug output before the line `const polyfillNode = debug(concat(new MergeTrees(inputNodes), {`

```
// chrome devtool
> needsFetchPolyfill
< false

> needsAbortControllerPolyfill
< false

> inputFiles
< (2) ["abortcontroller.js", "fetch.js"]
```

index.js
```diff
@@ -73,7 +74,10 @@ module.exports = {
       importTarget = this;
     }
 
-    app._fetchBuildConfig = app.options['ember-fetch'] || { preferNative: false };
+    const options = Object.assign({ preferNative: false, alwaysIncludePolyfill: true}, app.options['ember-fetch'] || {});
```
xg-wang: nit: The `|| {}` can be dropped

```js
app._fetchBuildConfig = Object.assign({
  preferNative: false,
  alwaysIncludePolyfill: true,
  browsers: this.project.targets && this.project.targets.browsers
}, app.options['ember-fetch']);
```

index.js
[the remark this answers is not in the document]
xg-wang: You can use [`allowNon`](https://github.com/broccolijs/broccoli-concat#advanced-usage), it will [allow the content set to empty string](https://github.com/broccolijs/broccoli-concat/blob/master/concat.js#L150).

The current way of checking these 2 booleans is producing wrong result, see my comment on first test result.

index.js
xg-wang: Yea agree you're right about the ergonomics here. Previously I thought `caniuse-api` was making HTTP requests but after some digging I realized they are packing the dataset with the package. So the result of `preferNative: true` should be only the enhancement of reducing the redundant asset.
To reflect the opt-out path for edge-cases you mentioned, can you also update the README about the new API?

mydea:
I pushed another set of changes, addressing the review. `alwaysIncludePolyfill` now defaults to `false`, and I've added a paragraph to the readme explaining the new polyfill semantics. I've also moved the option-reading up in the `treeForVendor` and pass the options to the `treeForBrowserFetch` method as argument. I also revised the test messages (I've just copied them from the preferNative tests and forgot to change them)

README.md
```diff
@@ -96,6 +96,8 @@ otherwise the polyfilled `fetch` will be installed during the first pass of the
 
 If set to `false`, the polyfilled `fetch` will replace native `fetch` be there or not.
 
+If all your browser targets (defined in `config/targets.js`) support native `fetch`, and `preferNative: true`, the polyfill will not be included in the output build. If, for some reason, you still need the polyfill to be included in the bundle, you can set `alwaysIncludePolyfill: true`.
```
xg-wang:
```suggestion
 If all your [browser targets](https://guides.emberjs.com/release/configuring-ember/build-targets/) support native `fetch`, and `preferNative: true`, the polyfill will not be included in the output build. If, for some reason, you still need the polyfill to be included in the bundle, you can set `alwaysIncludePolyfill: true`.
```

I think the first paragraph
> If set to `true`, the the fetch polyfill will be skipped if native `fetch` is available, ...

can just be replaced by your added paragraph.

mydea: Hmm, but that would then not cover the case that `preferNative: true`, but some targets do not have `fetch` support (which, for now, will probably be the case for many apps) - in that case, the setting will still have an effect (--> not use the polyfill if fetch is available).

PR #177, opened 2018-11-29
Only set default Content-Type header in adapter mixin

nlfurniss:
Came across some code I should have moved in 9917ba094ed94cb419144c690d5f4c1ea7216c59. 

Currently, every request that doesn't have a content-type header set is being set to `application/json` (to mimic logic in ED's adapter). This breaks the incoming logic from #174 where the browser needs to set the content-type header correctly for FormData.

### DIFF
--- a/addon/mixins/adapter-fetch.js
+++ b/addon/mixins/adapter-fetch.js
@@ -38,7 +38,25 @@ export default Mixin.create({
     if (adapterHeaders) {
       options.headers = assign(options.headers || {}, adapterHeaders);
     }
-    return mungOptionsForFetch(options);
+
+    const mungedOptions = mungOptionsForFetch(options);
+
+    // Mimics the default behavior in Ember Data's `ajaxOptions`, namely to set the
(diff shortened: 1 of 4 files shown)


xg-wang:
Sorry it's not very clear to me, is setting the `content-type` breaking `fetch` [setting](https://github.com/github/fetch/blob/v3.0.0/fetch.js#L233) correct `content-type` based on body type? This change only moved that part to the mixin, is it breaking for non-adapterMixin usage?

nlfurniss:
Yes, exactly. It is breaking for non-adapter mixin usage

xg-wang:
We can start moving [these](https://github.com/ember-cli/ember-fetch/blob/v6.2.2/assets/browser-fetch.js.t#L5) into polyfilled exports if we can ensure Fastboot parity.

```
      'FormData',
      'FileReader',
      'Blob',
      'URLSearchParams',
      'Symbol',
      'ArrayBuffer'
```

related PR https://github.com/ember-cli/ember-fetch/pull/76

PR #187, opened 2018-12-12
handle stringifying data that was created with Object.create(null)

meirish:
The new Object detection in 6.3.0 for checking when to stringify data does not work for objects that don't have a prototype. I didn't think this would be a common case, but it turns out we ran into it when using Ember's `hash` helper - I reconstructed the issue in this twiddle:

https://ember-twiddle.com/865249c0847b7465f53f0d025bab95aa?openFiles=templates.application.hbs%2C

### DIFF
--- a/addon/utils/mung-options-for-fetch.js
+++ b/addon/utils/mung-options-for-fetch.js
@@ -26,9 +26,10 @@ export default function mungOptionsForFetch(_options) {
     } else {
       // NOTE: a request's body cannot be a POJO, so we stringify it if it is.
       // JSON.stringify removes keys with values of `undefined` (mimics jQuery.ajax).
+      // If the data has no constructor we assume it is a POJO created with Object.create(null).
       // If the data is not a POJO (it's a String, FormData, etc), we just set it.
       // If the data is a string, we assume it's a stringified object.
-      if (options.data.constructor === Object) {
+      if (!options.data.constructor || options.data.constructor === Object) {
         options.body = JSON.stringify(options.data);
       } else {
         options.body = options.data;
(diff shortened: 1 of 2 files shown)


meirish:
I'm not sure why travis is failing on this change - looks like something related to FastBoot which I didn't think these changes should affect.

addon/utils/mung-options-for-fetch.js
```diff
@@ -26,9 +26,10 @@ export default function mungOptionsForFetch(_options) {
     } else {
       // NOTE: a request's body cannot be a POJO, so we stringify it if it is.
       // JSON.stringify removes keys with values of `undefined` (mimics jQuery.ajax).
+      // If the data has no constructor we assume it is a POJO created with Object.create(null).
       // If the data is not a POJO (it's a String, FormData, etc), we just set it.
       // If the data is a string, we assume it's a stringified object.
-      if (options.data.constructor === Object) {
+      if (!options.data.constructor || options.data.constructor === Object) {
```
nlfurniss: Are there cases where `data` doesn't have a constructor but isn't a POJO?

another option is to use `Object.prototype.toString.call(options.data) === "[object Object]"`, which works for your case

xg-wang: Might be better to use `Object.prototype.toString.call(options.data) === '[object Object]'`

meirish: I don't _think_ so, but happy to change it!

meirish: Updated!

meirish:
Thanks!

nlfurniss:
@meirish thanks for this! Released in 6.3.1

Issue #290, opened 2019-05-02
Enforce top-level add-on requirement


dnalagatla:
Ember-fetch requires the add-on to be a top-level dependency for an application to provide support for Fastboot. This is documented [here](https://github.com/ember-cli/ember-fetch#top-level-addon).

Ember-fetch should enforce the above requirement during the build time of an application if an add-on is using it as a transitive dependency. Recently, ember-data added a dependency on ember-fetch and if the application didn't include ember-fetch dependency, then fastboot failed to load the fetch module.

dnalagatla:
cc @nlfurniss @xg-wang for visibility.

PR #292, opened 2019-05-03
Added a check to enforce top-level dependency #290.

dnalagatla:
This PR adds a check to enforce top-level dependency requirement for ember-fetch addon.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,6 +68,16 @@ module.exports = {
   included: function() {
     this._super.included.apply(this, arguments);
 
+    let isApp = !this.project.isEmberCLIAddon();
+
+    let hasEmberFetch = !!this.project.findAddonByName('ember-fetch');
+    let hasEmberCliFastboot = !!this.project.findAddonByName('ember-cli-fastboot');
+
+    if(isApp && hasEmberCliFastboot && !hasEmberFetch) {
+      throw new Error(`Ember fetch is not installed as top-level dependency of the application using fastboot. Add ember-fetch as dependecy in application's package.json.
+      For details check here - https://github.com/ember-cli/ember-fetch#top-level-addon`);
+    }
+
     let app = this._findApp();
     let importTarget = app;
 


index.js
```diff
@@ -68,6 +68,15 @@ module.exports = {
   included: function() {
     this._super.included.apply(this, arguments);
 
+    let isApp = !this.project.isEmberCLIAddon();
+
+    let hasEmberFetch = this.project.addons.map(addon => addon.name).includes('ember-fetch');
```
stefanpenner: This looks good, but there is a more concise API that could be used:

`project.findAddonByName('ember-fetch')` https://github.com/ember-cli/ember-cli/blob/807f56e2e434f716757de02fd1f2c3251dadaf6f/lib/models/project.js#L634

I noticed this method was listed as private, but it's intent was to be public so I've fixed that:https://github.com/ember-cli/ember-cli/pull/8621

index.js
```diff
@@ -68,6 +68,15 @@ module.exports = {
...
+
+    if(isApp && !hasEmberFetch) {
+      throw new Error(`Ember fetch is not installed as top-level dependency of the application. Add ember-fetch as dependecy in application's package.json.
```
stefanpenner: @rwjblue should we start as a warning, and do a major version release that turns this into an error, or...

index.js
stefanpenner: Another approach, would be to do this work in the `init()` would be via `this.project.dependencies['ember-fetch']`

This approach would allow us to disable nested `ember-fetch` instances via `isEnabled() { }` hook, whereas if `isEnabled` returns false, we won't hit the `included` hook.

index.js
xg-wang: Yea I think we should follow deprecation workflow, error in next major (7.0.0).

index.js
```diff
@@ -68,6 +68,15 @@ module.exports = {
   included: function() {
     this._super.included.apply(this, arguments);
 
+    let isApp = !this.project.isEmberCLIAddon();
```
xg-wang: We won't need to detect app/addon, since for addon development `ember-fetch` should lives lin `devDependencies` it's also top-level.
You may intend to force error for Addon + dependencies.

xg-wang:
@dnalagatla @stefanpenner 
Just to clarify the hard requirement is for Fastboot support. In app without Fastboot, the problem of nested `ember-fetch` (or any addon with a custom vendor roll-up) is the first-write-win conflict.
@dnalagatla Can you also update the Readme, moving `top-level addon` from `Use with Fastboot` to upper level so it's no longer fastboot specific?

index.js
dnalagatla: I agree with that. what is the general flow that we follow for deprecation? What is the release branch we can update with warning?

index.js
dnalagatla: I think we need to detect if the project is an app or not. I want this error to be thrown during the application build only.

 Also, ember-fetch can be a transitive-dependency for add-on if the add-on has a dependency on another add-on which is dependent on ember-fetch. In that case it should not error. The top-level dependency requirement is only meant for application. 

Let me know if I missed anything here.

dnalagatla:
@xg-wang, I agree. I will update the readme to make this the requirement.

You are correct this issue is Fastboot specific and if apps are using fastboot, the ember-fetch as nested dependency fails to set up the fetch module. Apps not on fastboot, I think there should not be an issue. 

@stefanpenner @xg-wang 
Do you think we should check this a requirement if the application includes `ember-cli-fastboot` as a dependency. This way we can keep the text as it is in readme, as current text makes more sense.

index.js
dnalagatla: I made the change to use `findAddonByName` API for checking if addon exists. Also, updated the condition to check if the app has ember-cli-fastboot. If that addon exists and ember-fetch is not included throw the error.

dnalagatla: > Another approach, would be to do this work in the `init()` would be via `this.project.dependencies['ember-fetch']`
> 
> This approach would allow us to disable nested `ember-fetch` instances via `isEnabled() { }` hook, whereas if `isEnabled` returns false, we won't hit the `included` hook.

I am just wondering, using above and disabling nested fetch can have any side-effects.

xg-wang:
@dnalagatla Is this PR ready to merge? This requires a major bump that can go together with https://github.com/ember-cli/ember-fetch/pull/348.

index.js
xg-wang: We chatted about this. Using ember-fetch in a fastboot project will error anyway so this is not a deprecation, but rather making error more explicit.

PR #303, opened 2019-05-21
Add fetch response and error utils

BarryThePenguin:
Transitioning from `ember-ajax` to `ember-fetch` and noticed there weren't any [error detection helpers](https://github.com/ember-cli/ember-ajax#error-detection-helpers)

This PR adds similar helpers, mostly a copy/paste. Main motivation is to aid in transitioning between the two.

```js
import {
  isNotFoundResponse,
  isForbiddenResponse
} from 'ember-fetch/errors';
```

I looked through npm briefly to see if something similar already existed, but couldn't find anything. If we want to move forward with this, I can add some documentation to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,51 @@ If all your [browser targets](https://guides.emberjs.com/release/configuring-emb
 
 The way you do import remains same.
 
+### Error Handling
+
+`ember-fetch` comes with utility functions for matching responses and their response status codes.
+These utilities are to assist when transitioning from [`ember-ajax`](https://github.com/ember-cli/ember-ajax#error-handling)
+These include
+
+  - `isBadRequestResponse` (400)
+  - `isUnauthorizedResponse` (401)
+  - `isForbiddenResponse` (403)
+  - `isNotFoundResponse` (404)
+  - `isConflictResponse` (409)
+  - `isGoneResponse` (410)
+  - `isInvalidResponse` (422)
+  - `isServerErrorResponse` (5XX)
+  - `isSuccessResponse`
+
+And for aborted requests
+
+  - `isAbortError`
+
+
+```js
+import Route from '@ember/routing/route';
+import fetch from 'fetch';
+import {
+  isUnauthorizedResponse
+} from 'ember-fetch/errors';
+
+export default Route.extend({
+  model: function() {
+    const response = await fetch('/omg.json');
+    
+    if (response.ok) {
+      return response.json();
+    }
+    
+    if (isUnauthorizedResponse(response)) {
+      // handle 401 response
+      return;
+    }
+  }
+});
+
+```
+
 ## Browser Support
 
 * evergreen / IE10+ / Safari 6.1+ https://github.com/github/fetch#browser-support
--- a/README.md
+++ b/README.md
@@ -117,9 +117,11 @@ The way you do import remains same.
 
 ### Error Handling
 
-`ember-fetch` comes with utility functions for matching responses and their response status codes.
-These utilities are to assist when transitioning from [`ember-ajax`](https://github.com/ember-cli/ember-ajax#error-handling)
-These include
+A `fetch` response is successful if `response.ok` is true,
+otherwise you can read the status code to determine the bad response type.
+`fetch` will only reject with [network errors](https://fetch.spec.whatwg.org/#concept-network-error).
+
+`ember-fetch` provides some utility functions:
 
   - `isBadRequestResponse` (400)
   - `isUnauthorizedResponse` (401)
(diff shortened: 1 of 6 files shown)


addon/errors.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export function isFetchError(response: Response): boolean {
+  return response.ok === false
+}
+
+/**
+ * Checks if the given response represents an unauthorized request error
+ */
+export function isUnauthorizedError(response: Response): boolean {
+  if (isFetchError(response)) {
```
xg-wang: This doesn't seem necessary.

addon/errors.ts
```diff
@@ -0,0 +1,113 @@
...
+    return response.status === 410;
+  }
+
+  return false
+}
+
+/**
+ * Checks if the given response represents an "abort" error
+ */
+export function isAbortError(response: Response): boolean {
```
xg-wang: `AbortError` is from rejecting the `fetch` promise rather than setting status code to 0.
https://dom.spec.whatwg.org/#aborting-ongoing-activities

You may want to make a convenient method for `err.name == 'AbortError'`.

addon/errors.ts
```diff
@@ -0,0 +1,113 @@
...
+    return response.status >= 500 && response.status < 600;
+  }
+
+  return false
+}
+
+/**
+ * Checks if the given status code represents a successful request
+ */
+export function isSuccess(response: Response): boolean {
```
xg-wang: ember-ajax's `isSuccess` is for checking status code, do you think we should continue to make it checking status or response object?

addon/errors.ts
```diff
@@ -0,0 +1,113 @@
+/**
+ * Checks if the given response is a FetchError
+ */
+export function isFetchError(response: Response): boolean {
+  return response.ok === false
```
xg-wang: I think things are different in `fetch`.

`response.ok` is true for status 2xx, which will be false for 304 https://github.com/github/fetch/issues/241. I think we should remove this function and just use `isSuccess` you added at bottom.

Another thing to notice is that all the response check are successful in regards to network/abort and checking status from server. We also need to deal with `.catch` scenarios, where `response.type === 'error'` and `aborted flag` ([spec](https://fetch.spec.whatwg.org/#concept-response-type)).

xg-wang:
I think these are helpful to make people using fetch api easier, but there're some differences when migrating from Ajax to Fetch, I did a quick check on our polyfills (https://github.com/mo/abortcontroller-polyfill and https://github.com/github/fetch) my comments about response type are also true. Can you make the code changes and explicitly call out in Readme doc?

addon/errors.ts
BarryThePenguin: Similar to `isFetchError`, we probably don't need `isSuccess`?

```js
fetch().then(() => {
// is success
}).catch(() => {
// is error
})
```

xg-wang: Yea agree, it doesn't really provide that much value since it's just one liner `r => r.ok`...

Maybe rename `isXXXError` to `isXXXResponse`? And `isNetworkError` and `isAbortError
```
fetch().then((response) => {
    if (response.ok) {
        // is success
    } else  {
        // is xxx response
    }
}).catch((error) => {
        // is xxx error
})
```

BarryThePenguin:
Thanks for the review. It was very helpful. I learnt a fair bit reading the fetch and related specs

I've added a section to the readme too

README.md
```diff
@@ -115,6 +115,51 @@ If all your [browser targets](https://guides.emberjs.com/release/configuring-emb
 
 The way you do import remains same.
 
+### Error Handling
+
+`ember-fetch` comes with utility functions for matching responses and their response status codes.
+These utilities are to assist when transitioning from [`ember-ajax`](https://github.com/ember-cli/ember-ajax#error-handling)
```
xg-wang: These utils are useful for all fetch users not only migration from ember-ajax so I think this sentence is not necessary.

README.md
```diff
@@ -115,6 +115,51 @@ If all your [browser targets](https://guides.emberjs.com/release/configuring-emb
...
+These include
```
xg-wang: Can you explain more about fetch responses and errors? Maybe something like

---

A `fetch` response is successful if `response.ok` is true, otherwise you can read the status code to determine the bad response type. `fetch` will only reject with [network errors](https://fetch.spec.whatwg.org/#concept-network-error). 

`ember-fetch` provides some utils:

- `isBadRequestResponse` (400)
- `isUnauthorizedResponse` (401)
- `isForbiddenResponse` (403)
- `isNotFoundResponse` (404)
- `isConflictResponse` (409)
- `isGoneResponse` (410)
- `isInvalidResponse` (422)
- `isServerErrorResponse` (5XX)
- `isAbortError` [aborted network error](https://developers.google.com/web/updates/2017/09/abortable-fetch)

An example:

```js
import Route from '@ember/routing/route';
import fetch from 'fetch';
import {
  isAbortError,
  isUnauthorizedResponse,
  isServerErrorResponse
} from 'ember-fetch/errors';
export default Route.extend({
  async model() {
    try {
      let response = await fetch('/omg.json');
      if (response.ok) {
        return response.json();
      } else if (isUnauthorizedResponse(response)) {
        // handle 401 response
      } else if (isServerErrorResponse(response)) {
        // handle 5xx respones
      }
    } catch (error) {
      if (isAbortError(error)) {
        // handle aborted network error
      }
      // handle network error
    }
  }
});
```

addon/errors.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Checks if the given response represents a server error
+ */
+export function isServerErrorResponse(response: Response): boolean {
+  return response.status >= 500 && response.status < 600;
+}
+
+/**
+ * Checks if the given status code represents a successful request
+ */
+export function isSuccessResponse(response: Response): boolean {
```
xg-wang: Per our discussion can this be removed?

tests/unit/utils/determine-body-promise-test.js
```diff
@@ -1,4 +1,5 @@
 import { module, test } from 'qunit';
+import {Response} from 'fetch';
```
xg-wang: Ah do we have prettier setup for tests? It should be wrapped by whitespace.

tests/acceptance/error-test.js
```diff
@@ -0,0 +1,172 @@
+import { module, test } from 'qunit';
+import Pretender from 'pretender';
+import fetch, {AbortController} from 'fetch';
```
xg-wang: should be wrapped by white space.

README.md
xg-wang: I just checked the guide https://guides.emberjs.com/v3.10.0/routing/specifying-a-routes-model/#toc_fetch-example is using promise style, so I added 1 more commit to align with that.

BarryThePenguin: Ah great! Thanks for that. I think that's a good idea

PR #309, opened 2019-05-24
Add support for specifying a custom serialization function for query parameters.

edwinvanrijkom:
It's currently not possible to change how data gets mapped to query parameters. This can be cumbersome when a server may be expecting 'traditional style' formatted query parameters, for example.

With this PR  a custom serialization function can be specified.


xg-wang:
Thanks for the PR, closing as Rob suggested they may better lives in ember-data repo.

PR #348, opened 2019-08-16
Bump deps, drop node 6

xg-wang:


rwjblue:
Seems good to me, do we need to be worried about multiple major versions in play within the app at once anymore?

xg-wang:
@rwjblue I'm hoping to get https://github.com/ember-cli/ember-fetch/pull/292 in, then any fastboot app should be safe.

For multiple major in browser only app it's no different than other addons. Unfortunately [there's no perfect solution now](https://discuss.emberjs.com/t/how-should-we-resolve-conflicting-addons-and-dependencies/14523). 

The best way for app to guard is to install https://github.com/salsify/ember-cli-dependency-lint.

For addon authors if the addon depends on a specific version `ember-fetch` (or any other addon) functionality, there's always risk if not enforce a peer-dependency.

Issue #371, opened 2019-10-23
Sadness with ember-cli-fastboot


rwjblue:
Haven't had much time to dig into this yet, but I just spun up a new app with ember-cli@3.14.0-beta.1 and get an error when I add `ember-cli-fastboot`.

Steps to reproduce:

```
npx ember-cli@3.14.0-beta.1 new foo --yarn
cd foo
yarn add -D ember-cli-fastboot
ember serve --port 0
```

Then visit `/` in the app. I see:

```
Error: You must provide a hostWhitelist to retrieve the host
    at FastBootRequest.host (/Users/rjackson/src/rwjblue/test-disabling-loading-routes/node_modules/fastboot/src/fastboot-request.js:22:13)
    at Class._host (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/addon-tree-output/ember-cli-fastboot/services/fastboot.js:25:1)
    at Class.<anonymous> (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/addon-tree-output/ember-cli-fastboot/services/fastboot.js:30:1)
    at untrack (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/@ember/-internals/metal/index.js:3324:1)
    at untrack (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/@ember/-internals/metal/index.js:1247:1)
    at ComputedProperty.get (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/@ember/-internals/metal/index.js:3323:1)
    at Class.CPGETTER_FUNCTION [as host] (/var/folders/mt/yt2qjpl93ls98lrkrs81rzch000pzv/T/broccoli-130653rRm3Dk7C9Vf/out-261-broccoli_merge_trees/assets/@ember/-internals/metal/index.js:1044:1)
[...]

PR #372, opened 2019-10-25
fix: throwing w/ fresh ember-cli-fastboot serve

xg-wang:
ember-cli-fastboot requires configuring host whitelist before accessing
host from request. In ember-fetch, the error should be lazily thrown.

fix: #371 cc @rwjblue @pzuraq


xg-wang:
I think we can. https://github.com/ember-fastboot/ember-cli-fastboot/blob/v2.2.1/addon/services/fastboot.js#L14 seems `request` is safe to be stored.

xg-wang:
@rwjblue Will there be any memory leak concern for keeping the `request`? https://github.com/ember-fastboot/ember-cli-fastboot/blob/v2.2.1/addon/services/fastboot.js#L24

rwjblue:
No, I don't think so.

xg-wang:
Seems some strange floating dep caused `ember test` to fail.

Issue #391, opened 2019-11-18
Deprecation for ember data mixin for older ED versions


mydea:
After https://github.com/ember-cli/ember-fetch/pull/308, we get a lot of deprecation warnings in our app:

```
DEPRECATION: FetchAdapter is deprecated, it is no longer required for ember-data>=3.9.2 [deprecation id: deprecate-fetch-ember-data-support]
```

We cannot easily update ED (for other reasons), so we are currently using 3.4.x. 

The deprecation warning incorrectly suggests that we should stop using the FetchAdapter. I guess the deprecation should only be shown for users with ember-data >= 3.10?

Turbo87:
I'm seeing the same issue on https://github.com/rust-lang/crates.io

/cc @xg-wang

xg-wang:
@stefanpenner https://github.com/ember-cli/ember-fetch/pull/447 will be contradicting the request here to only log deprecation for newer ember-data versions. If we're going to drop the mixin, answer to this issue is to [manually silence the deprecation](https://api.emberjs.com/ember/release/functions/@ember%2Fdebug/registerDeprecationHandler) until ember-data upgrade is finished.

@mydea @Turbo87 what do you think?

mydea:
I guess that's OK, maybe not ideal but workable!

PR #392, opened 2019-11-18
Only show FetchAdapter deprecation for ember-data >= 3.9.2

mydea:
As described in #391 , the deprecation message for the FetchAdapter should only be shown if using Ember Data >=3.9.2.

### DIFF
--- a/addon/mixins/adapter-fetch.ts
+++ b/addon/mixins/adapter-fetch.ts
@@ -17,6 +17,7 @@ import {
   AjaxOptions
 } from 'ember-fetch/types';
 import { Fix } from '@ember/object/-private/types';
+import { lte } from 'ember-compatibility-helpers';
 
 /**
  * Helper function to create a plain object from the response's Headers.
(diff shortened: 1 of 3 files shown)


addon/mixins/adapter-fetch.ts
```diff
@@ -65,7 +66,9 @@ export default Mixin.create<FetchAdapter, DS.RESTAdapter>({
 
   init() {
     this._super(...arguments);
-    deprecate('FetchAdapter is deprecated, it is no longer required for ember-data>=3.9.2', false, {
+    // TS currently fails on the `lte()` check due to https://github.com/pzuraq/ember-compatibility-helpers/issues/34
+    // @ts-ignore
+    deprecate('FetchAdapter is deprecated, it is no longer required for ember-data>=3.9.2', lte('ember-data', '3.9.1'), {
```
rwjblue: Since its possible to use ember-data with the `ember-data` package, we should probably also check for `lte('@ember-data/adapter', '3.9.1')`.

mydea: Not 100% sure, but I think the packages were only introduced in 3.11? So I guess if you use packages, this should always be false anyhow? (Not sure how `lte()` handles the addon being missing at all?)

rwjblue: Ya, good point. We should test 😸

mydea: Turns out, this will not work :grimacing: As ember-compatibility-helpers throws an error at build time if trying to check a package that is not installed. E.g.:

```
Expected "ember-fetch" to have "@ember-data/adapter" as a dependency, but it was not found.
```

So I guess this approach will not work :(

mydea:
So turns out `ember-compatibility-helpers` will throw an error at build time if trying to check for an addon that is not installed. This means that the approach in this PR will not work... If anybody can think of another way to do that, that would be great.

addon/mixins/adapter-fetch.ts
runspired: I just used semver directly for this in a recent case. You can combine that with debug-macros if you need a build time flag but likely just a broccoli-funnel to remove the adapter if you need

xg-wang:
Closing per discussion in https://github.com/ember-cli/ember-fetch/issues/391

PR #399, opened 2019-11-24
Add section to the README for deprecations

locks:
I put it in the README to make it the easiest to find and because I wanted to get the suggestion down before I forgot, but I am completely open to alternative suggestions :)
I had to do some spelunking because the deprecation mentions `FetchAdapter` but the documentation names the default export as `AdapterFetch` to match the module name.


xg-wang:
LGTM.

re: `FetchAdapter` / `AdapterFetch` Ugh this is annoying, I may have followed ember-data's adapter naming and forgot to check the actual name.

Issue #398, opened 2019-11-20
Thoughts on an ember-ajax compatible service?


rwwagner90:
👋 

We are working on removing jQuery from our app, and thus moving from ember-ajax to ember-fetch, but we like a lot of the API of ember-ajax. Would you guys be interested in providing a compatible service that works with the ember-ajax API here or should that be a separate addon?

burritoIand:
@rwwagner90 I created a barebones service that ~matches ember-ajax. It could be enhanced to add the other http aliases etc

rwwagner90:
@burritoland we actually released an addon that is a full port of ember-Ajax to fetch. It’s https://github.com/expel-io/ember-ajax-fetch

burritoIand:
very nice, should we close this?

rwwagner90:
Yep, closing.

Issue #330, opened 2019-06-28
Rename import path to `ember-fetch`


chriskrycho:
It's a best practice in general, for imports to come from their package name rather than some more "convenient" name, and will become more so [as Embroider lands](https://github.com/emberjs/rfcs/blob/embroider/text/0000-embroider-v2-package-format.md#renaming-rules):

> New addons *should not* use renaming rules because it's confusing when the imports people type don't align with their real dependencies.

The import style here should simply be:

```js
import fetch from 'ember-fetch';
```

This would be a breaking change, but should also be trivially codemod-able.

rwjblue:
We could start by providing both modules, and deprecate the `fetch` one. Wouldn't need to be a breaking change until we drop support (which could happen "far in the future").

chriskrycho:
Yeah, I like that as a path forward.

ef4:
Given that global `fetch` is a stable WHATWG standard, what do we gain by keeping an importable ember-specific package, as opposed to polyfilling the global `fetch` in each environment we care about (just fastboot and IE11, I think) and declaring that all Ember apps and addons are free to use global `fetch`?

Implementation-wise, the two are equally interceptable.

chriskrycho:
Seems perfectly reasonable to me. We’d just want a nice configurable switch for that if it’s ending up in the vendor bundle, or (better?) to deliver it separately.

rwjblue:
I think thats fine in general, it would also allow easier "use native" (since we would literally do nothing if your targets all have `fetch`).  

We'd have to confirm that its "fine" in SSR environments too (since there **is not** a `fetch` global there), it would add to the list of expected globals in the `fastboot` sandbox setup (which is probably fine, but worth mentioning).

stefanpenner:
> Given that global fetch is a stable WHATWG standard, what do we gain by keeping an importable ember-specific package, as opposed to polyfilling the global fetch in each environment we care about (just fastboot and IE11, I think) and declaring that all Ember apps and addons are free to use global fetch?

@ef4 Im perfectly good with this, I prefer that approach.

Questions:

* In this world, does ember-fetch provide said polyfill or something else
* Currently `import fetch from 'fetch';` is always the polyfill, to ensure that 100% compatibility between ie11. This is clearly not ideal, but the alternatives come with some risks. My thought was, once i11 is off our radar we re-eval
[...]

Issue #222, opened 2019-02-17
beforeSend


miguelcobain:
Seems that some addons like ember-simple-auth make use of the `beforeSend` option. This option works on jquery ajax: http://api.jquery.com/jQuery.ajax/

Here, for example: https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/data-adapter-mixin.js#L81

Would it make sense for AdapterFetch to try to use this option?

At the moment, I don't think it is possible to use ember-simple-auth + ember-data + ember-fetch?

xg-wang:
It's also used in ember-data [JSONAPIAdapter](https://github.com/emberjs/data/blob/v3.7.0/addon/adapters/json-api.js#L181), and lots of [other addons](https://emberobserver.com/code-search?codeQuery=beforeSend). 

`ajaxOptions` in the fetch adapter doesn't call super, so other addon's mixins and ember-data's hash w/ `beforeSend` won't be mixed into `ajaxOptions`. I'm not sure whether this should happen in the adapter level or we provide such a hook in a `fetchService`. @rwjblue @nlfurniss Do you have any idea about this?

nlfurniss:
@xg-wang other `ajaxOptions` are def called (the adapter mixin has to be in the correct order). 

`beforeSend` isn't necessary here since the addon could change the options passed into `ajaxOptions`. However, `$.prefilter` doesn't have any support yet. I'm writing an RFC for how we want to handle `prefilter` in a jQuery-less world. Keep an eye out :-)

PR #447, opened 2020-02-14
Remove ember-fetch/mixins/adapter-fetch ember-fetch/ajax, drop node 8

stefanpenner:
ember-fetch provides alot of additional functionality, that functionality comes at several costs:

* size
* complexity / maintenance

I would like us to deprecate, and stop providing these functionality. This functionality is:

* ember-fetch/ajax: as jQuery is not longer supported, users should use fetch API’s
* ember-fetch/errors: we recommend folks use HTTP status codes directly
* ember-fetch/mixins/data: ember-data no longer needs this, it uses the fetch module provided by this add-on directly.

It allow users to utilize these features, we can point them to the 7.x branch, which we can maintain for a time. If users really want some of these features, other addons can be created, but having a kitchen sink of stuff here is unfortunately
For users who want these features, they can use [7.x branch](https://github.com/ember-cli/ember-fetch/tree/v7.x)

- [x] update readme
- [x] https://github.com/ember-cli/ember-fetch/tree/v7.x
- [ ] ????
- [x] consensus


nlfurniss:
One issue with removing the `utils` is that consuming apps then have to create all that logic themselves when making one-off `fetch` calls

stefanpenner:
@nlfurniss which are the ones that you find very useful for apps? I’m open to keeping stuff around, if it really provides value. :)

nlfurniss:
```
ember-fetch/utils/determine-body-promise
ember-fetch/utils/mung-options-for-fetch
ember-fetch/utils/serialize-query-params
```

I would argue that `serialize-query-params` is the least necessary, though it is convenient not to need to dupe this logic across multiple apps.

`determine-body-promise` is the most useful, as it makes dealing with the response very easy.

stefanpenner:
@nlfurniss in your opinion does it make sense for those utilities to be in this repo, or as their own dependency?

xg-wang:
## Ajax
Those 3 util functions are helpful for existing Ajax style requests, so their existence depends on the migration path we recommend for people using Ajax.
I think we should use https://github.com/ember-cli/ember-cli-version-checker#hassingleimplementation, with existing check added in https://github.com/ember-cli/ember-fetch/pull/292 it should be okay for other addons list `ember-fetch` as dependency, so if people find direct `fetch` isn't enough they can create their own wrapper.
Related: https://github.com/ember-cli/ember-fetch/issues/398 https://github.com/ember-cli/ember-fetch/issues/222

## Error
I think they are handy, if you feel strongly about it not being a pure polyfill I'm okay removing it.

## ember-data
+1 to removal as the deprecation has been added. However we need to patch the old major to fix https://github.com/ember-cli/ember-fetch/issues/391
My first try was blocked at https://github.com/pzuraq/ember-compatibility-helpers/pull/42, maybe I can do what runspired suggested to build a debug macro. But anyway that's not against removing it.

## Only provide fetch module
Have you looked at related proposal to not even have a fetch module? https://github.com/ember-cli/ember-fetch/issues/330

stefanpenner:
@nlfurniss / @xg-wang it sounds like we can split this into two steps:

* drop ember-data related mixins, as they are no longer needed for modern ember-data, but leave the utils + ajax helper. (land soon)
* explore the future of a more "pure" fetch world (explore and maybe land someday)

Thoughts?

xg-wang:
Talked offline we're going to drop mixin + bump node in v8 major bump and explore the pure fetch in the future.

bmaehr:
@xg-wang If ajax is not supported anymore (I just installed this module because of this and I'm really annoyed about this 'We remove jquery everywhere but have no alternative for doing ajax request") then you shuild remove it from the Q&A of the README (main page of the project).

xg-wang:
@bmaehr ah thanks for pointing out! Done in https://github.com/ember-cli/ember-fetch/pull/517

Issue #477, opened 2020-03-30
Operation outputFileSync is not allowed to use


Turbo87:
After trying to update from v7.1.0 to v8.0.0 we started to see the error below when trying to build our app:

```
Build Error (SimpleConcatConcat)
Operation outputFileSync is not allowed to use. Allowed operations are readFileSync,existsSync,lstatSync,readdirSync,statSync,writeFileSync,appendFileSync,rmdirSync,mkdirSync,unlinkSync,symlinkOrCopySync,symlinkSync,utimesSync
=================================================================================
ENV Summary:
  TIME: Mon Mar 30 2020 08:19:01 GMT+0000 (Coordinated Universal Time)
  TITLE: ember
  ARGV:
  - /usr/local/bin/node
  - /builds/qonto/qonto-js/node_modules/.bin/ember
  - exam
  - --split=8
  - --parallel
  EXEC_PATH: /usr/local/bin/node
  TMPDIR: /tmp
  SHELL: /bin/bash
  PATH:
  - /tmp/yarn--1585556330820-0.7482114818713204
  - /builds/qonto/qonto-js/node_modules/.bin
  - /usr/local/share/.config/yarn/link/node_modules/.bin
  - /builds/qonto/qonto-js/node_modules/.bin
  - /usr/local/share/.config/yarn/link/node_modules/.bin
  - /usr/local/libexec/lib/node_modules/npm/bin/node-gyp-bin
  - /usr/local/lib/node_modules/npm/bin/node-gyp-bin
  - /usr/local/bin/node_modules/npm/bin/node-gyp-bin
  - /usr/local/sbin
  - /usr/local/bin
  - /usr/sbin
  - /usr/bin
  - /sbin
  - /bin
  PLATFORM: linux x64
  FREEMEM: 7696748544
  TOTALMEM: 16222683136
  UPTIME: 2352
  LOADAVG: 2.4150390625,1.15625,0.99267578125
  CPUS:
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
  - Intel(R) Xeon(R) Platinum 8124M CPU @ 3.00GHz - 3000
[...]

PR #478, opened 2020-03-30
Bump broccoli-concat to pick up transitive dependency fix

stefanpenner:
cc @xg-wang / @SparshithNR


xg-wang:
Fixes https://github.com/ember-cli/ember-fetch/issues/477

PR #493, opened 2020-04-10
Resolve FastBoot dependencies at compile time

bobisjan:
This change makes FastBoot dependencies required at compile time, hence no need to `cd dist && npm install` as before. It somehow aligns with `ember-auto-import`, when app imports modules from 3rd packages, then EAI bundles these modules at compile time. I assume that this will be improved with Embroider in the future.

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,8 +143,50 @@ module.exports = {
   treeForPublic() {
     const fastbootEnabled = process.env.FASTBOOT_DISABLED !== 'true'
       && !!this.project.findAddonByName('ember-cli-fastboot');
-    return !this.parent.parent && fastbootEnabled
-      ? this._super.treeForPublic.apply(this, arguments) : null;
+
+    const include = !this.parent.parent && fastbootEnabled;
+
+    if (!include) {
+      return null;
+    }
+
+    let abortcontrollerNode = debug(new Rollup(path.dirname(path.dirname(require.resolve('abortcontroller-polyfill'))), {
+      rollup: {
+        input: 'src/ponyfill.js',
+        output: {
+          file: 'node-abortcontroller.js',
+          name: 'NodeAbortcontroller',
+          format: 'amd',
+          amd: {
+            id: 'node-abortcontroller',
+          }
+        }
+      }
+    }), 'node-abortcontroller');
+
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -143,8 +143,50 @@ module.exports = {
...
+        input: 'lib/index.js',
+        output: {
+          file: 'node-fetch.js',
+          name: 'NodeFetch',
+          format: 'iife'
+        }
+      }
+    }), 'node-fetch');
+
+    fetchNode = debug(map(fetchNode, (content) => `define('node-fetch', ['exports'], function(exports) {
```
bobisjan, line 178: I'm not sure if there is a better way to do this?

ef4:
While I agree with the high-level goal here, I don't think this code belongs in individual addons like ember-fetch. I see from [your comment on fastboot](https://github.com/ember-fastboot/fastboot/issues/251#issuecomment-612587368) that you already understand the need to fix this upstream and see this as a temporary step. In general, adding a whole new build pipeline (rollup in this case) to an addon makes it harder for us to control and optimize the whole build, so I usually discourage it.

You could work around this in your own app instead, with the bundling handled by ember-auto-import, something like this:

```js
// app/routes/application.js
...
fastboot: service(),
async beforeModel() {
  if (this.fastboot.isFastboot) {
     // we're choosing not to use Fastboot.require because we want all dependencies to get
     // bundled at build time instead. So here we provide the dependencies that addons would
     // have tried to Fastboot.require.
     let dependencies = await Promise.all([
       import('node-fetch'),
       import('abortcontroller-polyfill/dist/cjs-ponyfill'),
     ]);
     Fastboot.require = function(which) {
        switch (which) {
           case 'node-fetch':
              return dependencies[0].default;
           case 'abortcontroller-polyfill/dist/cjs-ponyfill':
              return dependencies[1].default;
           default:
              throw new Error(`we did not bundle: ${which}`);
        }
     }
  }
}
...
```

In this case, ember-auto-import will always see and bundle the dependencies but they will go into their own lazy chunk(s) that never load in the browser.

bobisjan:
Closing in favour of https://github.com/ember-cli/ember-fetch/issues/656#issuecomment-1173152715

PR #553, opened 2020-08-11
[V7.x] Fix whatwg-fetch 3.4 global object name change compatibility

xg-wang:
Backport https://github.com/ember-cli/ember-fetch/pull/552. There're folks cannot upgrade because old ember-data

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ jobs:
       script:
         - yarn lint:js
         - yarn test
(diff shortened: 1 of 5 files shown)


xg-wang:
> error semver@7.3.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"

okay I give up \_(:з」∠)\_

PR #568, opened 2020-10-06
Updates the type for data property to be less restrictive.

LuisAverhoff:
Removes the restriction of the data property having to be type of PlainObject where only the properties are strings, objects with strings or an array of objects containing strings. Allow objects will null or undefined properties


xg-wang:
This looks reasonable, I dug a bit down, the final place the type is used is at

https://github.com/ember-cli/ember-fetch/blob/v8.0.2/addon/utils/serialize-query-params.ts#L56-L67
and the expected type indeed can be as the PR specifies

```ts
declare function encodeURIComponent(uriComponent: string | number | boolean): string;
```

Issue #600, opened 2021-03-19
Ember 3.27+ issues deprecation for accessing `window.Ember`


rwjblue:
The code here:

https://github.com/ember-cli/ember-fetch/blob/eaf18b64fd86ac57a55861874d460a10a4ba151c/assets/browser-fetch.js.t#L4

Accesses `window.Ember`, which on Ember 3.27 and higher issues a deprecation.

rwjblue:
I think the best fix for this (that would support both older Ember and newer) would be to change our `.transpileTree` invocation to actually transpile modules for us:

https://github.com/ember-cli/ember-fetch/blob/eaf18b64fd86ac57a55861874d460a10a4ba151c/index.js#L124-L128

Then we'd change our `assets/browser-fetch.js.t` to just use ES6 imports for Ember.

Issue #337, opened 2019-07-04
Embroider compat


ef4:
Opening this issue because @xg-wang asked in Discord:

> Hi @ef4 , do you know what should be changed for https://github.com/ember-cli/ember-fetch to be embroider compatible?

There are a couple different ways to answer.

First, does `ember-fetch` work today under `@embroider/compat`? Yes, as far as I can tell it works flawlessly, at least in the browser environment. We are still working through fastboot support issues in embroider, so I didn't test `ember-fetch` in the fastboot environment.

Second, how can ember-fetch be implemented as a native V2-formatted addon (assuming https://github.com/emberjs/rfcs/pull/507 is accepted)?

I think it gets simpler to implement using the macro system. All the custom broccoli code becomes unnecessary.  Here's a simplified sketch to show the basic idea:

```js
// package.json
{
  ...
  "ember-addon": {
    "version": 2,
    "renamed-packages": {
      "ember-fetch": "fetch"
    },
    "build": "build.js"
  },

  // we have an optional peer dependency on ember-cli-fastboot
  // because we want to be able to check its configuration.
  "peerDependencies": {
    "ember-cli-fastboot": "*"
  },
  "peerDependenciesMeta": {
    "ember-cli-fastboot": {
      "optional": true
    }
  }
}
```

`index.js` would look something like:

```js
import { macroCondition, getOwnConfig, getConfig, importSync } from '@ember/macros';

let implementation;

if (macroCondition(getConfig('ember-cli-fastboot').isFastboot)) {
  // we end up here if ember-cli-fastboot is installed and
  // if we're actually running inside fastboot. This assumes
  // that ember-cli-fastboot will expose an `isFastboot`
  // property to the macro system, but that is a safe assumption.
  // We can make sure it does so.
  implementation = importSync('node-fetch').default;
} else {
  if (macroCondition(getOwnConfig().needsPolyfill)) {
    // this is where you can put the shimming code
    // that currently lives in browser-fetch.js.t. I simplified
    // it here for brevity.

    // this library sets the global fetch
    importSync('whatwg-fetch');
  }
  implementation = fetch;
}
export default implementation;
```

and there would be a `build.js` that looks like:

```js
export default class {
  configure() {
    return {
      needsPolyfill: checkIfPolyfillNeeded()
    }
  }
}
```

[...]

PR #620, opened 2021-06-01
Fix ember global

chancancode:
A minimal fix for #600. We should probably follow through with the refactor suggested by @rwjblue in #600 but that can be done later.

This is probably not very embroider friendly, but neither is the current code. It's probably easier to just rewrite the shims in terms of ES modules and embroider macros, i.e. what was suggested in #337.

### DIFF
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -1,5 +1,6 @@
 (function (originalGlobal) {
   <%= moduleHeader %>
+    var Promise = RSVP.Promise;
     var supportProps = [
       'FormData',
       'FileReader',
--- a/assets/browser-fetch.js.t
+++ b/assets/browser-fetch.js.t
@@ -1,7 +1,5 @@
 (function (originalGlobal) {
-  define('fetch', ['exports'], function(exports) {
-    'use strict';
-    var Promise = originalGlobal.Ember.RSVP.Promise;
+  <%= moduleHeader %>
     var supportProps = [
       'FormData',
       'FileReader',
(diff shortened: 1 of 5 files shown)


assets/browser-fetch.js.t
```diff
@@ -1,7 +1,14 @@
 (function (originalGlobal) {
-  define('fetch', ['exports'], function(exports) {
+  if (!require.has('ember')) {
+    define('ember', ['exports'], function(exports) {
```
rwjblue, line 3: Hmm, this seems not great. Defining the module could break other code / addons trying to sniff the module their own purposes. 🤔

rwjblue: Since we are templating this file, we could tell it if we are on modules ember or not (either adding the `ember` module as a dep or falling back to global ember automatically).

rwjblue (approved):

izelnakri:
Hi @chancancode , 

> This is probably not very embroider friendly, but neither is the current code. It's probably easier to just rewrite the shims in terms of ES modules and embroider macros, i.e. what was suggested in https://github.com/ember-cli/ember-fetch/issues/337.

I had a similar problem when upgrading this package/dependency for `mber`. `mber` does bunch of hacks to get this dependency included in the fastboot environment, now I need to adjust the hack again for `<%=` templates. 

Despite `fetch` coming to newer versions of node.js, I'll probably still need this package for runloop awareness and its helpers functions. So I'd really appreciate if we can rewrite these shims in terms of ES modules and macros ***as soon as possible*** instead of relying on a broccoli dependent temporary solution.

PR #737, opened 2023-10-04
Rename import path to 'ember-fetch'

lflo5727:
As per this issue: https://github.com/ember-cli/ember-fetch/issues/330 `ember-fetch` is incompatible with v2 addons and still required for some Ember projects. 

This fixes the issue with v2 addons and follows more direct naming conventions. 
I also removed the TS portion from the README as that workaround is no longer needed now that the module name matches the import name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,21 +37,6 @@ Available imports:
 import fetch, { Headers, Request, Response, AbortController } from 'ember-fetch';
 ```
 
-### Use with TypeScript
-To use `ember-fetch` with TypeScript or enable editor's type support, You can add `"fetch": ["node_modules/ember-fetch"]` to your `tsconfig.json`.
-
-```json
-{
-  "compilerOptions": {
-    "paths": {
-      "fetch": [
-        "node_modules/ember-fetch"
-      ]
-    }
-  }
-}
-```
-
 ### Use with Ember Data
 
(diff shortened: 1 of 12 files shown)


lflo5727:
It looks like the CI tests are using too new of a version of Node for this project. They are on 18 and the project needs 14 at newest it seems

NullVoxPopuli:
> They are on 18 and the project needs 14 at newest it seems

would you be willing to help out with some infra updates?
I've made a to-do list here: https://github.com/ember-cli/ember-fetch/issues/738

lflo5727:
> > They are on 18 and the project needs 14 at newest it seems
> 
> would you be willing to help out with some infra updates? I've made a to-do list here: #738
> 

Anything in particular? Should we make this a v2 addon here?
Our project I'm wanting to use this for is still stuck on v3 of Ember at the moment, so not super keen on dropping support for everything < 4 in this work.

NullVoxPopuli:
> Anything in particular? Should we make this a v2 addon here?

imo, all of it needs to be done, including v2 conversion (which is probably ~ 99% tests, and 1% actual code)

> so not super keen on dropping support for everything < 4 in this work.

we'd only really be dropping IE11, I've updated #738 to call that out specifically

PR #740, opened 2023-10-05
Rewrite ember-fetch to a v2 addon

StephanH90:
## Description
The goal is to make `ember-fetch` a v2 addon, removing all the old fetch polyfill stuff while maintaining backwards compatibility. This MR is the first step in this process.

This MR basically:

- removes all the broccoli magic stuff
- uses native fetch
- exposes a `setupFetchWaiter` function which provides backwards compatibility in tests
- continues to export a `fetch` from `ember-fetch` for backwards compatibility (its just the native fetch)

## Breaking
The only change that has to be made, if somebody wants to upgrade to `v9` is that they have to import a `setupFetchWaiter` function which wraps the native fetch with an `waitForPromise`. This means that fetch will continue to work in tests as it has done in the past. Using the `setupFetchWaiter` would look something like this:

```javascript
import { setupFetchWaiter } from 'ember-fetch';

setupFetchWaiter();
```

In development and production builds this does nothing.

## Next steps
In the next steps various other addons will have to be pulled into test apps to make sure that they continue to work with this version of fetch.


NullVoxPopuli:
Looks like nohe modules got committed 🙃

README.md
NullVoxPopuli, line 1: We'll want to keep README.md and RELEASE.md, `.release-it.json`, and the CHANGELOG.md should be move in to the `ember-fetch` directory.

.github/dependabot.yml
NullVoxPopuli, line 1: we probably want to keep this as well

ember-fetch/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+    isSetup = true;
+  };
+} else {
+  // do nothing if we're not testing.
+  setupFetchWaiter = () => {};
+}
+
+export {
+  setupFetchWaiter,
+  fetch, // fetch is re-exported here for backwards compatbiility
```
NullVoxPopuli, line 39: this `fetch` will be undefined unless `isTesting`, which isn't quite right.
We'll want it to be `globalThis.fetch`

NullVoxPopuli:
Can you re-target main/master? I closed the `next` branch, and made a v8 for bugfixes/backports

PR #741, opened 2023-11-01
chore: removes deprecated assign polyfill

Pixelik:
## Chore: makes addon Ember-5-compatible by removing deprecated `assign` polyfill


Pixelik:
ping @nlfurniss 🙏 🙂

addon/utils/mung-options-for-fetch.ts
```diff
@@ -13,7 +12,7 @@ import {
 export default function mungOptionsForFetch(
   options: AjaxOptions
 ): FetchOptions {
-  const hash = assign(
+  const hash = Object.assign({},
```
void-mAlex, line 15: I think we may need to drop support for ember < 4 to do this as 3.28 still supports IE 11

Pixelik: would you like to me to drop support for ember < 4 in this PR ?

mansona: Is that true 🤔 this just moves out of the concern of the addon and could be handled by a [babel plugin](https://babeljs.io/docs/babel-plugin-transform-object-assign) or something

mansona: essentially it would be bad if we dropped support for 3.28 for no reason

Pixelik: ember-fetch does not work on Ember 5 without this change

mansona: sorry I wasn't 100% clear 🙈 we should make this change, but we shouldn't stop testing in Ember 3.28

Pixelik: I've updated the code.
I had to remove support for `Ember < 3.28` because of `ember-qunit` but that's about it.
So this PR now supports Ember >= 3.28 (including Ember 5).
I've refactored the code so that neither the `ember assign polyfill` nor `Object.assign()` are used in production code. So IE is also still supported.

Pixelik:
All tests are passing except for **5** **mocha** tests failing during `yarn test:node` 🤷‍♂️

PR #743, opened 2023-11-09
update to v3.28 with ember-cli-update

mansona:


void-mAlex:
hasn't this already been done on https://github.com/ember-cli/ember-fetch/pull/741 ?

Pixelik:
It hadn't actually, I moved @mansona's addition from this PR into my PR recently actually 🙂 
Now all tests are passing in my PR except for the Node ones but this is kind of unknown territory for me, seems to be TS related.. 😕

tehhowch:
A benefit of this PR vs #741 is that it isn't a breaking change, so apps which are not yet on node 16 (for whatever reason) can still update

PR #744, opened 2024-10-06
Duplicate of #732 to see how CI reacts

Windvis:


Windvis:
Nevermind, it seems I never contributed before.

PR #754, opened 2024-11-05
Convert to v2, addon. Just wrap `fetch`.

NullVoxPopuli:
This PR is (what I think is) the goal. 
We can decide if we want to move towards this goal incrementally or deciding that this is what we want and skip all the in-between.

To move incrementally, we'd need to resolve the following PRs:
- https://github.com/ember-cli/ember-fetch/pull/753
- https://github.com/ember-cli/ember-fetch/pull/743
- https://github.com/ember-cli/ember-fetch/pull/742
- https://github.com/ember-cli/ember-fetch/pull/741
- https://github.com/ember-cli/ember-fetch/pull/732

Which... we can do -- I'm not sure it's worth keeping CI green through all that though. `ember-fetch` is in a bit of disrepair.

Support:
- 3.28+ (easily), maybe 3.4+
- ember-auto-import @ v2+ or embroider
- `@ember/test-waiters` v3+
- any node (node doesn't matter anymore)
- continue the renamed-modules support (import from fetch)

I'm not sure what the fastboot support story would be, but Node natively has `fetch`, `AbortController`, etc.

Dropped Support:
- Configuration: PREFER_NATIVE, NATIVE_PROMISE
- jQuery AJAX
- You no longer import AbortController. `AbortController` exists on `globalThis`
- You no longer import Headers. `Headers` exists on `globalThis`
- You no longer import Response. `Response` exists on `globalThis`

Since `ember-auto-import` tests against 3.4.0, and ember-auto-import is the only real requirement for this change, there is **no change** in supported ember-source.

If anything, the constraining function is `@ember/test-waiters`

Changes:
- [x] delete the old addon
- [x] create new fetch export
- [x] copy over old utils as separate exports
  - [x] remove './types', because these exist all natively
  - [x] remove './utils/determine-body-promise' because it used jQuery AJAX options
  - [x] remove './utils/mung-fetch-options' because it used jQuery AJAX options
- [x] **no build -- folks can just point at git to test this**
- [x] separate test app
- [x] yarn -> pnpm

After this PR:
- Add release plan
- remove release-it


src/utils/serialize-query-params.js
```diff
@@ -1,17 +1,17 @@
...
-
 const RBRACKET = /\[\]$/;
 
+function isPlainObject(obj) {
+  return Object.prototype.toString.call(obj) === '[object Object]';
+}
+
 /**
  * Helper function that turns the data/body of a request into a query param string.
  * This is directly copied from jQuery.param.
```
bertdeblock, line 9: There's also https://github.com/knowledgecode/jquery-param, so wondering if we should still include this in ember-fetch?

NullVoxPopuli:
Todo:
- remove ember-fetch from blueprints

esbanarango:
This is great!

src/index.js
```diff
@@ -0,0 +1,27 @@
...
+          return waitForPromise(parsePromise);
+        };
+      }
+
+      return original;
+    },
+  });
+}
+
+export default wrappedFetch;
```
mansona, line 27: I mean why not just add the default to line 3 😂

NullVoxPopuli: My brain doesn't like it haha

src/index.js
```diff
@@ -0,0 +1,27 @@
...
+    get(target, prop, receiver) {
+      let original = Reflect.get(target, prop, receiver);
+
+      if (['json', 'text', 'arrayBuffer', 'blob', 'formData'].includes(prop)) {
+        return (...args) => {
+          let parsePromise = original(...args);
+
+          return waitForPromise(parsePromise);
+        };
+      }
```
fivetanley, line 20: This didn't work in Safari for our application. It turns out that's because some properties use "internal slots" for property access, so you have to wrangle the `this` value to be the right thing. I lifted the implementation from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#no_private_property_forwarding

```suggestion
      let original = target[prop];

      if (typeof original === 'function') {
        return function (...args) {
          let maybePromise = original.apply(this === receiver ? target : this, args);
          // only wait for promise if it is a promise
          if (maybePromise && typeof maybePromise.then === 'function') {
            return waitForPromise(maybePromise);
          }
          
          return maybePromise;
        };
      }
```

Since there are other properties like `toString` etc which can get accessed, I made it only `waitForPromise` if we have a promise-like return value as well.

fivetanley: The conditional use of waitForPromise makes it so it only gets called on promises. if waitForPromise is called with a non-promise, it seems to error and hang the test suite with a waiter that never resolves.

void-mAlex: rfc has it as 
```js
let parsePromise = original.call(target, ...args);
```
which I've successfully used

NullVoxPopuli:
The RFC landed on 

https://github.com/emberjs/ember-test-waiters/blob/master/addon/src/wait-for-fetch.ts

So I'm going to close this and open a deprecation pr on the readme of ember fetch

